PR #25, opened 2018-05-10
fix(publish): encode the git tag in the release notes post so that ta…

### LINKED ISSUE #26: Encode git tag in publish release notes URI
Hey all,

Apologies for the sloppiness of this PR, doing this in a meeting: https://github.com/semantic-release/gitlab/pull/25

I couldn't get @semantic-release/gitlab to publish correctly from our lerna monorepo for shared modules.

I found the issue was with the git tags having a `/` in them that wasn't URI encoded and as such, causing a 404.

Hopefully what I've done is good enough to merge. 🙂

Thanks,
Christos.

constantology:
…gs that contain a forward slash `/` — e.g. a lerna monorepo that publishes to an npm scope — don't cause a 404

-----

Fix #26

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -8,13 +8,14 @@ module.exports = async (pluginConfig, {options: {repositoryUrl}, nextRelease: {g
   const {gitlabToken, gitlabUrl, gitlabApiPathPrefix} = resolveConfig(pluginConfig);
   const repoId = getRepoId(gitlabUrl, repositoryUrl);
   const apiUrl = urlJoin(gitlabUrl, gitlabApiPathPrefix);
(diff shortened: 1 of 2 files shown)


constantology:
Hey all,

Associated issue here: https://github.com/semantic-release/gitlab/issues/26

pvdlg:
Thanks for the PR. Looks good!
For info there is no need to open both an issue and a PR, as a PR is by itself an issue.
If you have question about a potential implementation or a complex feature you can open an issue first so we can discuss there. If it's a simple change as here, you can just open the PR, no need for the issue.

constantology:
@pvdlg thanks so much. 🙂

Sorry for the double up, I'll keep it to one or the other, as per your advice, in future. 😁

PR #41, opened 2018-08-16
Re-write usage phrase in the README.md

pproenca:
In the usage section,  'Semantic-release configuration file' was mentioned as the place to write the configuration. This phrase might mislead people into thinking that there is an additional file, other than `package.json` for writing the configuration for this tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ authentication is supported.
 
 ### Usage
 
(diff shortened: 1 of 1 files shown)


pvdlg:
This is intentional, semantic-release does support multiple type of configuration files. See [configuration](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/configuration.md#configuration).

PR #79, opened 2019-06-29
feat(assets): add asset links to release using api

rdaniels6813:
and remove them from markdown.

I noticed when trying to use assets that were uploaded to Gitlab using this plugin that they are not available through the API unless I want to try parsing the markdown field for them.

Since Gitlab exposes an API for adding links to a release, and those links are included under the default display for assets inside of Gitlab, I thought it might be an improvement to add them using the API, and removing the markdown section.

# Before 
image

# After
image


pvdlg:
Is this API a new feature? I'm pretty sure when we added the list of asset to the markdown that wasn't available?

Can you check what is the oldest version of GitLAb were this new API exists?
If we break compatibility with older GitLAb version we need to make a major release.

rdaniels6813:
Introduced in gitlab 11.7

https://docs.gitlab.com/ee/api/releases/links.html

rdaniels6813:
Are you willing to do a major version bump so that this can be supported?  What do you need from my end to move forward?

nfantone:
Looking forward to this being released!

Issue #94, opened 2019-10-23
Test suite improvements


esciara:
Problems like #92 where not caught by the unit tests. They need to be improved.

pvdlg:
See https://github.com/semantic-release/gitlab/pull/95#issuecomment-546760455

PR #95, opened 2019-10-23
test(verify): improving tests on verify see #94

esciara:
Part of #94 .

Work concentrated on improving test cases where an exception is thrown.

 - check Error type thrown to be that AggregateError
 - check that only the expected error is thrown
 - added assert messages to inform which of the asserts failed

Only test not modified is  "Throw SemanticReleaseError for invalid repositoryUrl", as I met some issues with it. Will be correcting it in subsequent work.

### DIFF
--- a/test/verify.test.js
+++ b/test/verify.test.js
@@ -1,6 +1,8 @@
 import test from 'ava';
 import nock from 'nock';
 import {stub} from 'sinon';
+import AggregateError from 'aggregate-error';
+import {HTTPError} from 'got';
 import verify from '../lib/verify';
 import authenticate from './helpers/mock-gitlab';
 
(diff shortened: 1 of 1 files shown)


test/verify.test.js
```diff
@@ -307,77 +309,129 @@ test.serial('Verify "assets" is an Array of Object with a glob Arrays in path pr
...
     verify(
       {assets},
-      {env, options: {repositoryUrl: 'https://gitlab.com/semantic-release/gitlab.git'}, logger: t.context.logger}
-    )
+      {env, options: {repositoryUrl: `https://gitlab.com/${owner}/${repo}.git`}, logger: t.context.logger}
+    ),
+    AggregateError,
+    'Error thrown but of unexpected type.'
   );
+  const errors = [...aggregateError];
```
pvdlg: The intermediate `aggregateError` seems unnecessary. Why not just get the `errors` directly?

test/verify.test.js
```diff
@@ -307,77 +309,129 @@ test.serial('Verify "assets" is an Array of Object with a glob Arrays in path pr
...
 
-  t.is(error.name, 'SemanticReleaseError');
-  t.is(error.code, 'EINVALIDASSETS');
+  const expectedNumOfErrors = 1;
```
pvdlg: What's the point of adding the variable `expectedNumOfErrors`? It's only used on the next line and it's different for each test. I'd rather just pass the number directly to the test.

I prefer to have test that are simple to read and understand. Adding a bunch intermediate variable just create more code, more possibilities for a mistake and more cognitive load to read the tests.

test/verify.test.js
```diff
@@ -307,77 +309,129 @@ test.serial('Verify "assets" is an Array of Object with a glob Arrays in path pr
   t.true(gitlab.isDone());
 });
 
-test('Throw SemanticReleaseError if "assets" option is not a String or an Array of Objects', async t => {
+test.serial('Throw SemanticReleaseError if "assets" option is not a String or an Array of Objects', async t => {
+  const owner = 'test_user';
```
pvdlg: Why adding those intermediate variables?

test/verify.test.js
```diff
@@ -307,77 +309,129 @@ test.serial('Verify "assets" is an Array of Object with a glob Arrays in path pr
...
+    AggregateError,
+    'Error thrown but of unexpected type.'
   );
+  const errors = [...aggregateError];
 
-  t.is(error.name, 'SemanticReleaseError');
-  t.is(error.code, 'EINVALIDASSETS');
+  const expectedNumOfErrors = 1;
+  t.is(errors.length, expectedNumOfErrors, `Expected ${expectedNumOfErrors} error but got ${errors.length}.`);
+  t.is(errors[0].name, 'SemanticReleaseError', 'Unexpected error name.');
```
pvdlg: I feel the message `Unexpected error name.` is stating the obvious at beat and misleading at worst.
It's not that the error name is wrong, it's that a different error has been thrown.
It felt to me that the verification is so simple and the `ava` output does a really good job at indicating the assertion that failed and provide the correct/wrong values that adding a message here wasn't necessary.

test/verify.test.js
esciara: Because it is less readable. Adding the intermediate helps understanding the code better.

test/verify.test.js
esciara: Actually I added the intermediate to make it more readable and easier to understand, and to keep the DRY principle. (It is used in the assert and in the assert error message => you don't have to replace it at two places => less bugs). Even though it is one more constant it is more expressive : it is what it says it is : the expected number of errors. Without it you would have to guess it. So I am using this approach to actually lower the cognitive load.

esciara: Actually I added the intermediate to make it more readable and easier to understand, and to keep the DRY principle. (It is used in the assert and in the assert error message => you don't have to replace it at two places => less bugs). Even though it is one more constant it is more expressive : it is what it says it is : the expected number of errors. Without it you would have to guess it. So I am using this approach to actually lower the cognitive load.

test/verify.test.js
pvdlg: I'm sorry but I don't see how making the reader go through more hops to understand something is better in any way...

test/verify.test.js
pvdlg: The added message is not really useful. It just state the obvious. It's number comparison and `ava` out the expect line and the value it was passed too. Adding a sentence to state that number `x` is different than number `y` doesn't aadd any value vs `ava` output other than having to write more code.

esciara: Just to explain my point (and after you consider it, since you are the maintainer, I will follow whatever rules you see fit):

Here is the `ava`'s output without the failure message (I put the expected to `2` just to make it fail):
image

image

And here is the same with the failure message:
image

image

Without the message, you have no clue of what is going and have to go to the detailed failure report, to read the code, to try to understand what it means to finally figure out what the problem is about.

With the message,  you immediately have an explanation next to the test in the initial summary, and for every test failing. No need to dig in: you have an overview in the summary of why each test has failed. You then can go to the detailed failure report if need be to get more insight.

Now, the message itself night not have been the smartest, and could be changed to something like  "Only one error should have been thrown but more have." And this way, no need for an `expectedNumOfErrors`. But I do see an advantage in having a message.

What do you think?

test/verify.test.js
esciara: Just to make sure I get it right.

You prefer this:
```
  const errors = [
    ...(await t.throwsAsync(
      verify(
        {assets},
        {env, options: {repositoryUrl: `https://gitlab.com/${owner}/${repo}.git`}, logger: t.context.logger}
      ),
      AggregateError,
      'Error thrown but of unexpected type.'
    )),
  ];

```  

Over this?:
```
  const aggregateError = await t.throwsAsync(
    verify(
      {assets},
      {env, options: {repositoryUrl: `https://gitlab.com/${owner}/${repo}.git`}, logger: t.context.logger}
    ),
    AggregateError,
    'Error thrown but of unexpected type.'
  );
  const errors = [...aggregateError];
```

test/verify.test.js
esciara: The reason I added failure message is explained about. Waiting for you feedback on the principle. 

As for the message itself, yes, I agree it is not the best one... What about something like `Unexpected error thrown` or `Error other than SemanticReleaseError thrown`?

image

test/verify.test.js
esciara: I added the `gitlab` because if you don't create it like for other test you get an extra `EINVALIDGLTOKEN` error, hence your test is not accurate (not checking only the `EINVALIDASSETS` error code is thrown. So I copied the same approach than the `Verify token and repository ...` tests to make sure that no other errors would be thrown.

But is you have another way to do it, I can correct it to that way.

esciara:
Hi @pvdlg , I see that you have on-boarded some of the suggested changes of this PR in the release https://github.com/semantic-release/gitlab/releases/tag/v4.0.2 . Looking forward to your feedback on the remaining discussions.

pvdlg:
I added the additional checks to make sure only the expected errors are thrown. Not interested by the other changes. Thanks for contributing though!

PR #96, opened 2019-10-23
docs(README): add GitLab personal token scope mention

Nargonath:


pvdlg:
In some (most) circumstances the `write_repository` is also required.

Can rewrite it as the following?
```
Create a [personal access token](https://docs.gitlab.com/ce/user/profile/personal_access_tokens.html) with the `api` scope and make it available in your CI environment via the `GL_TOKEN` environment variable. If you are using `GL_TOKEN` as the [remote Git repository authentication](https://github.com/semantic-release/semantic-release/blob/master/docs/usage/ci-configuration.md#authentication) it must also have the `write_repository` scope.
```

Issue #92, opened 2019-10-21
EGLNOPERMISSION and EINVALIDGLTOKEN thrown even with a valid token that has the correct permissions


esciara:
Despite having set `GITLAB_TOKEN` with a valid token that has the correct permissions, I have the following error:
```
[8:33:25 AM] [semantic-release] [@semantic-release/gitlab] › ℹ  Verify GitLab authentication (https://gitlab.com/api/v4)
[8:33:26 AM] [semantic-release] › ✖  Failed step "verifyConditions" of plugin "@semantic-release/gitlab"
[8:33:26 AM] [semantic-release] › ✖  EGLNOPERMISSION The GitLab token doesn't allow to push on the repository root/my-project.
The user associated with the GitLab token (https://github.com/semantic-release/gitlab/blob/master/README.md#gitlab-authentication) configured in the GL_TOKEN or GITLAB_TOKEN environment variable must allows to push to the repository root/my-project.

Please make sure the GitLab user associated with the token has the permission to push (https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions) to the repository root/my-project.

[8:33:26 AM] [semantic-release] › ✖  EINVALIDGLTOKEN Invalid GitLab token.
The GitLab token (https://github.com/semantic-release/gitlab/blob/master/README.md#gitlab-authentication) configured in the GL_TOKEN or GITLAB_TOKEN environment variable must be a valid personal access token (https://docs.gitlab.com/ce/user/profile/personal_access_tokens.html) allowing to push to the repository root/my-project.

Please make sure to set the GL_TOKEN or GITLAB_TOKEN environment variable in your CI with the exact value of the GitLab personal token.

AggregateError: 
    SemanticReleaseError: Invalid GitLab token.
        at module.exports (/usr/local/lib/node_modules/@semantic-release/gitlab/lib/get-error.js:6:10)
        at module.exports (/usr/local/lib/node_modules/@semantic-release/gitlab/lib/verify.js:58:21)
    SemanticReleaseError: The GitLab token doesn't allow to push on the repository root/my-project.
        at module.exports (/usr/local/lib/node_modules/@semantic-release/gitlab/lib/get-error.js:6:10)
        at module.exports (/usr/local/lib/node_modules/@semantic-release/gitlab/lib/verify.js:67:19)
    at /usr/local/lib/node_modules/semantic-release/lib/plugins/pipeline.js:54:11
[...]

PR #100, opened 2019-10-26
fix: verify badly configured reverse proxy + more complete error mess…

### LINKED ISSUE #99: EMISSINGREPO thrown when in fact the reverse proxy in front of GitLab is incorrectly configured
See toward the end of https://github.com/semantic-release/gitlab/issues/63#issuecomment-527080331.

Verification currently does not make a distinction when the repository is really missing (response with `statusCode` at `404` and `statusMessage` is `Project Not Found`) and when there is a problem with a badly configured reverse-proxy (same but `statusMessage` is `Not Found`).

esciara:
…ages

Now throws EMISSINGREPO only when response with statusCode at 404 and statusMessage is "Project Not
Found" and throws EUNEXPECTED404 when there is a problem with a badly configured reverse-proxy
(similar response but with statusMessage is "Not Found"). Also adds information when error is thrown
to help with troubleshooting.

fix #99 and re #92

### DIFF
--- a/lib/definitions/errors.js
+++ b/lib/definitions/errors.js
@@ -13,22 +13,22 @@ module.exports = {
     )}) must be an \`Array\` of \`Strings\` or \`Objects\` with a \`path\` property.
 Your configuration for the \`assets\` option is \`${stringify(assets)}\`.`,
   }),
-  EINVALIDGITLABURL: () => ({
-    message: 'The git repository URL is not a valid GitLab URL.',
+  EINVALIDGITLABURL: ({gitlabUrl, repositoryUrl}) => ({
+    message: `The git repository URL "${repositoryUrl}" is not a valid URL for the GitLab instance at ${gitlabUrl}.`,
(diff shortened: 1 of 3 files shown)


pvdlg:
This case of 404 seems unrelated to this plugin or to GitLab and specific to your particular proxy configuration. I don't think such edge case should be handled by this plugin, especially because there is no guarantees that a 404 is a sign of that specific issue.
GitLab might return a 404 for different reasons (such as a repo private with a token that doesn't have access to it) and parsing the error message looking for the string `Project Not Found` seems very unreliable as GitLab can change their message at any time.

As a general rule if there is something that intercept the calls between the client and the server and modify it in a way that make the request invalid it's not the responsibility of the client to try to figure out what's going on and making guesses.
For example if the server cannot be reached we just display a message saying so and we don't try to figure out if the network interface is down, or if the DNS is not configured properly or if there is a routing problem.
We expect to have a working network between the client and the server and we expect that nothing alter the requests and make them invalid.

esciara:
Fair enough.

Issue #65, opened 2019-02-02
Publish links instead of files


joonhwa:
My project is building and releasing containers. How can i show links in the release notes? Currently it shows a list of files but is not ideal in my case as i rather have a link to an image in the container registry.

joonhwa:
hi.. my apologies.. i should have provided more details.

So my project is primarily concerned with building Docker images and hosting them from Gitlab Container Registry. The ci job now runs in 3 stages: build, test, tag. During the tagging stage i would like to automatically git tag that particular commit to make it as a release and also generate a release note. I managed to get everything running so far but what i wanted in the release note is to be able to create some text to point users to that new image in the Container Registry. In my case, my users do not really need to have the uploaded src files.

Hope this clarifies...

pvdlg:
Ok I understand. So the only solution is to create a custom `generateNotes` plugin that would append the link you want to add to the release note.

Alternatively you can do that with a script using the `@semantic-release/exec` plugin and setting the `generateNotesCmd` option to point to a script that would output the link you want to append to the release note.

joonhwa:
ok cool... i will give that a try.. thanks!

PR #119, opened 2019-12-26
feat: support external urls in assets

nfantone:
Allow `assets` to declare `{ url, name }` objects in order to produce links to external resources as an alternative to uploading files. Both `url` and `name` (and its alias, `label`) support templating through [`lodash.template`](https://lodash.com/docs/4.17.15#template) using a `{{}}` delimiter.

```json
{
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
      "gitlabUrl": "https://custom.gitlab.com",
      "assets": [
        {"path": "dist/asset.min.css", "label": "CSS distribution"},
        {"url": "https://s3.amazonaws.com/{{env.ASSETS_BUCKET}}", "label": "asset@{{nextRelease.version}}"},
      ]
    }]
  ]
}
```

This will help ease off situations where the artifact/s produced by a build are already uploaded to some other repository elsewhere in the pipeline (see https://github.com/semantic-release/gitlab/issues/65).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
+
+# We recommend you to keep these unchanged
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@ The plugin can be configured in the [**semantic-release** configuration file](ht
       "gitlabUrl": "https://custom.gitlab.com",
       "assets": [
         {"path": "dist/asset.min.css", "label": "CSS distribution"},
-        {"path": "dist/asset.min.js", "label": "JS distribution"}
+        {"path": "dist/asset.min.js", "label": "JS distribution"},
+        {"url": "https://s3.amazonaws.com/{{env.ASSETS_BUCKET}}", "label": "asset-{{nextRelease.version}}"},
       ]
     }],
   ]
@@ -66,9 +67,13 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
 |-----------------------|--------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `gitlabUrl`           | The GitLab endpoint.                                               | `GL_URL` or `GITLAB_URL` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `https://gitlab.com`. |
 | `gitlabApiPathPrefix` | The GitLab API prefix.                                             | `GL_PREFIX` or `GITLAB_PREFIX` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `/api/v4`.      |
-| `assets`              | An array of files to upload to the release. See [assets](#assets). | -                                                                                                                                                                       |
+| `assets`              | An array of files to upload or external urls to add to the release. See [assets](#assets). | -                                                                                                                                                                       |
 
-#### assets
+#### `assets`
+
+Assets can be either **files to be uploaded** to Gitlab and attached to the release or **custom URLs linking to external resources**. Both can be used and configured interchangeably using the `assets` configuration setting.
+
+##### File assets
 
 Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` of
 [globs](https://github.com/isaacs/node-glob#glob-primer) and `Object`s with the following properties:
@@ -84,19 +89,36 @@ can be a `String` (`"dist/**/*.js"` or `"dist/mylib.js"`) or an `Array` of `Stri
 
 If a directory is configured, all the files under this directory and its children will be included.
 
-**Note**: If a file has a match in `assets` it will be included even if it also has a match in `.gitignore`.
+> **Note**: If a file has a match in `assets` it will be included even if it also has a match in `.gitignore`.
 
-##### assets examples
+Examples:
 
-`'dist/*.js'`: include all the `js` files in the `dist` directory, but not in its sub-directories.
+- `'dist/*.js'`: include all the `js` files in the `dist` directory, but not in its sub-directories.
 
-`[['dist', '!**/*.css']]`: include all the files in the `dist` directory and its sub-directories excluding the `css`
+- `[['dist', '!**/*.css']]`: include all the files in the `dist` directory and its sub-directories excluding the `css`
 files.
 
-`[{path: 'dist/MyLibrary.js', label: 'MyLibrary JS distribution'}, {path: 'dist/MyLibrary.css', label: 'MyLibrary CSS
+- `[{path: 'dist/MyLibrary.js', label: 'MyLibrary JS distribution'}, {path: 'dist/MyLibrary.css', label: 'MyLibrary CSS
 distribution'}]`: include the `dist/MyLibrary.js` and `dist/MyLibrary.css` files, and label them `MyLibrary JS
 distribution` and `MyLibrary CSS distribution` in the GitLab release.
 
-`[['dist/**/*.{js,css}', '!**/*.min.*'], {path: 'build/MyLibrary.zip', label: 'MyLibrary'}]`: include all the `js` and
+- `[['dist/**/*.{js,css}', '!**/*.min.*'], {path: 'build/MyLibrary.zip', label: 'MyLibrary'}]`: include all the `js` and
 `css` files in the `dist` directory and its sub-directories excluding the minified version, plus the
 `build/MyLibrary.zip` file and label it `MyLibrary` in the GitLab release.
+
+##### External URL assets
+
+An `Array` of `Object`s with the following properties:
+
+| Property          | Description                                                                                                                                                                                              | Default             |
+| ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------- |
+| `url`             | **Required.** An absolute URL linking to an external resource.  | -                   |
+| `label` or `name` | Text of the anchor link on the release.                                                                                                                                                             | The value of `url`. |
+
+> All properties support interpolation of `context` variables through [`lodash.template`](https://lodash.com/docs/4.17.15#template), using `{{}}` as delimiter.
+
+Examples:
+
+- `[{url: 'https://assets.company.io/releases/latest', label: 'Latest Release' }]`: set a link to `https://assets.company.io/releases/latest` with the text _'Latest Release'_ in the GitLab release.
+
+- `[{url: '{{env.ASSETS_REPOSITORY_URL}}/releases/{{env.ARTIFACT_NAME}}.zip', label: '{{env.ARTIFACT_NAME}}-{{nextRelease.version}}.zip' ]`: interpolate variables present in `semantic-release` `context` to produce a link to a custom URL with a label.
--- a/lib/gitlab-api.js
+++ b/lib/gitlab-api.js
@@ -1,0 +1,50 @@
+const got = require('got');
+const urlJoin = require('url-join');
+
+function createGitlabApi(gitlabApiUrl, {token}) {
+  const apiOptions = {json: true, headers: {'PRIVATE-TOKEN': token}};
+
+  const links = async (repoId, gitTag, {name, url}) => {
+    const linksUrl = urlJoin(
+      gitlabApiUrl,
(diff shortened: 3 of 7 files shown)


pvdlg:
It seems this PR make a lot of formatting/style change and re-organization of the code.
As a general rule we won't accept a PR that both implement a feature and make some unrelated refactoring. We want to keep the PRs easy to review and have atomic commits so we can revert changes independently if necessary.
If you want to make changes in the code style/organization you should do so in another PR. But we won't accept a refactoring PR unless a very compelling is made. If it's a matter of preferences in style/tooling we won't accept the PR.

Regarding the feature itself, I'm not sure I understand the purpose of it. Correct me if I'm wrong, but it seems to add an asset to the release with a URL hardcoded in the config.

What's the point of adding a URL to the release that is always the same? If your delivered artifact is always accessible on the same URL why not just put that in your doc/README?
Also you would have all releases pointing to the same URL. So old releases would point to an artifact corresponding to a different version (the latest released).

If you want to have links to external artifacts in your GitLab release I would recommend to create a dedicated plugin that would: make the release of the artifact and then create the GitLab release with the unique URL of the artifact.

nfantone:
@pvdlg Thanks for the comments. Let me address some of them.

> It seems this PR make a lot of formatting/style change and re-organization of the code.

The only true "re-organization" of the code is the addition of a new `gitlab-api.js` module that encapsulates API calls related logic. Seemed fitting given that I needed to re-use bits of it. It also makes `publish` definition less cluttered, easier to read and encapsulates responsibilities. All core changes are circumscribed to `publish.js`. Nothing major, really. 

Everything else was done to simplify the addition of the new feature and integrate it without breaking stuff.

There are no formatting/style changes in here.

> Regarding the feature itself, I'm not sure I understand the purpose of it. Correct me if I'm wrong, but it seems to add an asset to the release with a URL hardcoded in the config.

It's explained and exemplified in both the PR description and the new `README.md`.   It enables users, like myself, to create Gitlab releases that look like this:

image

Where the URL of the link does _not_ point to a previously uploaded file - but rather, an external file, typically tied to some other build/upload pipeline (like Docker images, `.zip` artifacts, documentation or repositories). Right now, you don't support this use case.

You _can_ set a "hardcoded" URL, if you need to - but the typical use case here is to actually generate the link from the environment (both `name` and `url` support variables interpolation, such as `{{env.ASSETS_REPOSITORY_URL}}`).

> What's the point of adding a URL to the release that is always the same? If your delivered artifact is always accessible on the same URL why not just put that in your doc/README?
Also you would have all releases pointing to the same URL. So old releases would point to an artifact corresponding to a different version (the latest released).

It follows from the above that this ☝️  is just not the case.

> If you want to have links to external artifacts in your GitLab release I would recommend to create a dedicated plugin that would: make the release of the artifact and then create the GitLab release with the unique URL of the artifact.

While you _could_ you do that, this would also be true for every other piece of logic not supported by `semantic-release` or its plugins.

What I'm trying to do here is to _create a Gitlab Release_. This plugin already does that. And it's only a few changes away from supporting my use case (one that is not _that_ uncommon, I'd risk saying). This introduces no breaking changes and [others may benefit from them](https://github.com/semantic-release/gitlab/issues/65). It clearly belongs here, IMHO.

pvdlg:
> The only true "re-organization" of the code is the addition of a new gitlab-api.js module that encapsulates API calls related logic. Seemed fitting given that I needed to re-use bits of it. It also makes publish definition less cluttered, easier to read and encapsulates responsibilities. All core changes are circumscribed to publish.js. Nothing major, really.

Just through a quick look at the code changes:
- The `gitlab-api.js` doesn't seems to be related to this particular feature. It's just another level of encapsulation that in my opinion doesn't bring any simplification other than displacing some of the code. It does't factorise anything nor reduce the amount of code
- The `lib/publish.js` has been rewritten entirely and almost doesn't contain any off the original code. That doesn't seems related to this feature and is certainly major contrary to what you said.
- There is changes in the formatting of markdown
- There the addition of `.editorconfig`
- There is a change in the prettier config in `package.json`

> Everything else was done to simplify the addition of the new feature and integrate it without breaking stuff.
Didn't look at everything in details but many of those changes are completely unrelated to the feature implemented and should be in a different PR. But again we won't accept a refactoring PR unless a compelling argument is made. If it's a matter of personal preferences / opinion we probably won't merge it.  

> Where the URL of the link does not point to a previously uploaded file - but rather, an external file, typically tied to some other build/upload pipeline (like Docker images, .zip artifacts, documentation or repositories). Right now, you don't support this use case.
> You can set a "hardcoded" URL, if you need to - but the typical scenario here is to actually generate the link from the environment (both name and url support variables interpolation, such as {{env.ASSETS_REPOSITORY_URL}} or {{nextRelease.version}}).

Thanks for clarifying. However, I think it's a narrow implementation of the feature as it comes with several limitations:
- The link can be generated only with the info know by semantic-release, so if your external asset contains a non deterministic element (a date, a unique ID, etc...) the link cannot be re-created and there is no way to add this asset to the release
- It's prone to error as the users need to make sure two different things (what publish the asset and the config of the external URL in this plugin) end up generating the same URL

I think the `success` step is the proper way to handle such case.
Each `publish` plugin has the possibility to return a JSON object with the `url` and `name` property. Those informations are passed to the `success` plugins. For example tha't how the GitHub plugin generate comments like https://github.com/semantic-release/gitlab/pull/115#issuecomment-568135460 which contain the name and url of all the releases made by the `publish`s plugins. 

So we could implement a `success` step in the GitLab plugin that would:
- Make sure the `publish` has run and a GitLab release was created (by storing that in its local state)
- for each `context.releases` with a `url` and a `name` add that link to the GitLab  release

With this solution there wouldn't be a need to configure the external URLs, so none of the limitations I mentioned above would apply.

You would simply have to make sure the plugin you use to publish your artifact return a JSON object with `name` and `url`. That also mean 1 artifact per `publish` plugin. Not sure if that's a problem?

Would that cover your needs?
I'm willing to implement it this way if that works for you, as it also possibly involve changes in the core (at least to allow the plugin to exclude its own release from `context.releases`).

nfantone:
@pvdlg Thanks once again for your reply. I understand what you're saying. Let's see how we can roll this out.

> The `gitlab-api.js` doesn't seems to be related to this particular feature.

Well - like I explained before, I needed to be able to re-use logic that would trigger calls to Gitlab and the way `publish` was written made that tricky. And, frankly, it's already kind of messy by mixing logic from different layers that could use some separation of concerns.

But I understand your own concern, nonetheless. 

> The `lib/publish.js` has been rewritten entirely and almost doesn't contain any off the original code.

It has been re-written where it needed to be, in my humble opinion. How it was structured made it harder to extend and/or accommodate the new feature without breaking existing functionality.

> There is changes in the formatting of markdown.

I changed some headings that were using `#` to just plain text because the extra sub-heading introduced by the new `asset` type made it so sub-titles looked _smaller_ on GH than regular bold text - which was bad. And added bullet points to a list that didn't have them to match the piece of documentation I wrote. 

I can rollback those if you don't like them. That's fine.

> I think it's a narrow implementation of the feature as it comes with several limitations

You write "several" but list only a couple. And only one is actually true.

> The link can be generated only with the info know by semantic-release, so if your external asset contains a non deterministic element (a date, a unique ID, etc...) the link cannot be re-created 

This [ain't necessarily so](https://www.youtube.com/watch?v=vMYM75JFrIY). `semantic-release` has access to your entire environment. Take a look at the examples provided. You can set _any_ environment variable pointing to a "unique id" or date (or whatever you need) and use it in your URL via templating.

With a configuration such as:

```json
{
  "url": "https://assets.company.io/{{env.MY_RANDOM_UNIQUE_ID}}/asset-{{nextRelease.version}}",
  "name": "asset.{{env.MY_SECRET_DATE}}-{{nextRelease.version}}"
}
```

Executing:

```sh
# Retrieve STORED_DATE and unique STORED_ID from storage

MY_SECRET_DATE=${STORED_DATE:-`date +%m-%d-%Y`}
MY_RANDOM_UNIQUE_ID =${STORED_ID:-`openssl rand -base64 12`}

# Persist MY_SECRET_DATE and MY_RANDOM_UNIQUE_ID in storage

npx semantic-release
```

Will get you an asset pointing to `https://assets.company.io/mUGJc0gbZa3eo80p/asset-0.6.1` named `asset.12-27-2019-0.6.1`, every time.

There's no limitation to this.

> It's prone to error as the users need to make sure two different things (what publish the asset and the config of the external URL in this plugin) end up generating the same URL

I agree. This is actually true - but inevitable if the plugin itself is _not_ uploading the files... which is the use case we want to support. Also, this is going to be part of the same automatic build pipeline, so it's unlikely that "users" will have to "make sure". One part builds, uploads and sets relevant variables. `@semantic-release/gitlab` just reads those variables.

I'm open to ideas on this if you think it can be made more robust.

> So we could implement a `success` step in the GitLab plugin

I don't dislike the idea! Looks like it could blend into the `semantic-release` ecosystem very well. However...

> You would simply have to make sure the plugin you use to publish your artifact return a JSON object with name and url.

- You'd be making it harder for users out there _not_ using `semantic-release` for uploading or generating artifacts (a.k.a.: [non-JS, non-npm projects](https://github.com/semantic-release/semantic-release/blob/master/docs/support/FAQ.md#can-i-use-semantic-release-to-publish-non-javascript-packages)).

- You'd force the need for people to write, maintain and configure an _extra_ plugin. Maybe `exec` could come to the rescue here?

- You'll no longer have a single source of truth for configuring assets as _any_ plugin in the chain can populate your `context.releases` and force calls to Gitlab API (unless I'm not following you right).

> With this solution there wouldn't be a need to configure the external URLs, so none of the limitations I mentioned above would apply.

As per before, there *would* if you're not using `semantic-release` to generate the artifact. There would be no advantage, actually.

Thoughts?

pvdlg:
Refactoring the refactoring / code organization I'm going to state once more the point I made earlier and I won't debate it further:
- We won't accept a PR that both implement a feature and refactor the code more than directly required by the feature itself
- We won't accept any refactoring / code organization change unless a very compelling argument is made in advance. Opinions and preferences are not compelling arguments. It should be understood without saying that when it comes to opinions and preferences the ones of the maintainer prevails over the ones of irregular contributor.

Regarding the implementing, using `context.env` to share information is not a solution we are going to go with.
First, it wouldn't work as `context.env` is not shared across plugins, unless we merge https://github.com/semantic-release/semantic-release/pull/1391. This PR is under consideration for a very specific and unrelated situation. It might not be merged.

Secondly, as in any plugin system, it's important to not create dependencies between plugins as that would be error prone and confusing. We want each plugin to always be usable on their own independently of other plugins. That means plugins should not share informations with each other, and should not rely on another plugin to have done something.

Here is the answer to some of your comments.

> This would dispel your fears of being "prone to error", wouldn't it?
For you yes as you are aware of this works internally.
However the immense majority of users do not know the inner working of semantic-release. They want a tool that works without have to understand every implementation details. If you write a paragraph in the doc explain how 2 config lines have to match each other, we will get dozen of but report saying it's broken and dozen asking why one would have to configure the same thing twice. 

> You'd be making it harder for users out there not using semantic-release for uploading or generating artifacts (a.k.a.: non-JS, non-npm projects).

I don't see that as a problem. semantic-release is build with the assumption that it is making the publish of the release. That's why we have a plugin system and the `@semantic-release/exec`, so we can support any type publication for a release. The scenario of using semantic-release while making the release with something else outside off semantic-release is not something we will ever support as it just defeat the purpose of semantic-release.

> You'd force the need for people to write, maintain and configure an extra plugin. Maybe exec could come to the rescue here?
Wouldn't you solution with `context.env` do exactly that? Wouldn't it force every plugin to write a specific variable in `context.env`?

`@semantic-release/exec` does support returning a JSON object in the `publish` step by printing the  JSON string to the console:
```bash
// do the release
// then
echo '{"url":"http://my-release.com","name":"My Release"}'
```
You solution with `context.env` will not work with `@semantic-release/exec` as there is no way in the executed script to modify `context.env`.

Plugin author already have an incentive to return this object as it is standardized by semantic-release and will currently be used by any plugin implementing the `success`. While there is no reason for a plugin author to write a specific variable in `context.env` to have their plugin work with a specific use case that uses this specific variable.
Basically the `publish` return is standardized and generic, while `context.env.MY_VERY_SPECIFIC_VARIABLE` is neither.

> You'll no longer have a single source of truth for configuring assets as any plugin in the chain can populate your context.releases and force calls to Gitlab API (unless I'm not following you right).

Isn't it exactly what your solution with `context.env` does? Any plugin can set any variable that might or might not trigger something on `@semantic-release/gitlab`.

The return value of `publish` is actually the single source of truth. Each `publish` make the publication of one asset in one place (npm package, GitHub release, GitLAb release etc...) and is responsible to set the `url` and `name`. So you have one release object per publish plugin, set once by that plugin. How is that not the exact definition of single source of truth?

The role of the `publish` step is to publish and the role of `success` is to do something when a release is successful. In that case when the release is successful the `@semantic-release/gitlab` plugin would add the published asset links to the GitLab release metadata, similarly to the way the `@semantoc-release/github` plugin add a comment in each issue/PR included in the release to add the links to those releases.

pvdlg:
Closing as no response was provided

Issue #156, opened 2020-07-09
Support `CI_JOB_TOKEN` auth


dosuken123:
GitLab CI injects `CI_JOB_TOKEN` to allow a pipeline job to access a project resource through public v4 API. Not all of APIs support job token auth, however, [release creation API](https://docs.gitlab.com/ee/api/releases/#create-a-release) is already supported. So in theory, executing `GITLAB_TOKEN=$CI_JOB_TOKEN` should be suffice to run semantic-release.

However, as I tested https://gitlab.com/dosuken123/semantic-release-test/-/jobs/628942282, it didn't succeed as semantic-release tried [remote Git repository authentication](https://github.com/semantic-release/gitlab#gitlab-authentication). I'm not sure why it's necessary as the tag creation is done via the above API call.

If `CI_JOB_TOKEN` auth is possible, it's very convenient to run semantic-release in GitLab CI as users don't need to create [PAT](https://github.com/semantic-release/gitlab#gitlab-authentication).

mmuenker:
For me, the `CI_JOB_TOKEN` works fine.

**GitLab**: 13.4.0-pre
**GitLab Runner**: 13.3.1 (shared runner)

**.gitlab-ci.yml**
```yaml
stages:
- release

release:
  needs: []
  image: registry.gitlab.com/rxap/gitlab-ci/semantic-release
  stage: release
  script: semantic-release
  variables:
    GITLAB_TOKEN: $CI_JOB_TOKEN
  interruptible: true
  rules:
  - if: '$CI_COMMIT_REF_NAME == "master"'
    when: on_success
```

mattkasa:
@mmuenker I'm not able to reproduce a working release using your example with `$CI_JOB_TOKEN`, I get:

```
[9:15:28 PM] [semantic-release] › ✖  EGITNOPERMISSION Cannot push to the Git repository.
```

Can you think of anything different in your set up that would cause it to work?

virtuman:
is it possible that it was meant to work in gitlab 13.4.* and not 13.3 ?

EdwinSmulders:
I've tried this but I can't find any circumstances that made this work like @mmuenker says is possible. However, I also cannot find why it doesn't work - it seems like it fails before the gitlab plugin even does anything.

secustor:
I have looked into this. At the moment I don't see how this could work at the moment, at least for `/api/v4/`.  
[...]

PR #162, opened 2020-09-01
fix(resolve-config.js): use the env CI_JOB_TOKEN for the gitlab token

mmuenker:
This merge would resolve issue #156.

If the env `GL_TOKEN` and `GITLAB_TOKEN` are not defined the env `CI_JOB_TOKEN` would be used.

With the usage of the env `CI_JOB_TOKEN` the user who triggers the pipeline would also be the author of the new tag and release. This could cause some permission problems if the user who triggers the pipeline has not permission to create tags or/and releases.

### DIFF
--- a/lib/resolve-config.js
+++ b/lib/resolve-config.js
@@ -15,6 +15,7 @@ module.exports = (
       GITLAB_URL,
       GL_PREFIX,
       GITLAB_PREFIX,
+      CI_JOB_TOKEN,
     },
   }
 ) => {
@@ -31,7 +32,7 @@ module.exports = (
       : 'https://gitlab.com');
 
   return {
-    gitlabToken: GL_TOKEN || GITLAB_TOKEN,
+    gitlabToken: GL_TOKEN || GITLAB_TOKEN || CI_JOB_TOKEN,
     gitlabUrl: defaultedGitlabUrl,
     gitlabApiUrl:
       userGitlabUrl && userGitlabApiPathPrefix


MRigal:
@mmuenker Although tests pass, this won't work as the CI_JOB_TOKEN does not have push rights on the repository to push the tag. Only release creation is permitted. See https://docs.gitlab.com/ce/api/#gitlab-ci-job-token

Here an example of output using CI_JOB_TOKEN

'EGITNOPERMISSION',
  details: '**semantic-release** cannot push the version tag to the branch `semantic-release` on the remote Git repository with URL `https://gitlab-ci-token:[secure]@dev.jonasundderwolf.de/db/db-react-ui`.\n' +

EvanCarroll:
@MRigal While not ideal, I don't see why that should hold up this patch. Currently there is nothing stopping you from _explicitly_ providing a token that doesn't have access to the repository. Why should we stop a fall back from the same thing? 

Also the [GItLab Release API creates the tag for you if `tag_name` doesn't exist and ref is set.](https://docs.gitlab.com/ee/api/releases/index.html#create-a-release) It would seem this would be easy to do, I think we would [just have add `ref`](https://github.com/semantic-release/gitlab/blob/25a9011fd31eac326ff9796dfd9ca680326f4788/lib/publish.js#L88)

MRigal:
@EvanCarroll you may read the original issue https://github.com/semantic-release/gitlab/issues/156

Since the Gitlab Job Token does not allow to push, it's not possible to push version changes or tags to the repo. So the PR is a nice try but pointless, as Gitlab is too restrictive here

gdraganic:
@MRigal It isn't pointless in case of SSH auth to the repository which handles writing to repo and pushing tags. If `CI_JOB_TOKEN` could be used for auth to Releases API, Personal Access Token can be avoided. This gives us two security benefits:

1. SSH key only has write access to the project repository
2. There is no long living PAT with full api and repository:write access

Also worth mentioning that Gitlab API auth uses `JOB-TOKEN` header instead of `PRIVATE-TOKEN`.

EvanCarroll:
> @MRigal It isn't pointless in case of SSH auth to the repository which handles writing to repo and pushing tags. If `CI_JOB_TOKEN` could be used for auth to Releases API, Personal Access Token can be avoided. This gives us two security benefits:
> 
> 1. SSH key only has write access to the project repository
> 2. There is no long living PAT with full api and repository:write access
> 
> Also worth mentioning that Gitlab API auth uses `JOB-TOKEN` header instead of `PRIVATE-TOKEN` for auth with `CI_JOB_TOKEN`.

My fork supports this, fyi.

flickerfly:
Looks like there is more activity around this idea in #1729.
https://github.com/semantic-release/semantic-release/issues/1729#issuecomment-1354472875

PR #184, opened 2020-12-08
refactor: replace deprecated Tags API call

### LINKED ISSUE #183: Migrate from deprecated Tags API endpoint to Release API endpoint
In GitLab 14.0, the ability to manage release data (i.e. release notes) through the `/projects/<project id>/repository/tags/<tag name>/release` endpoint will be removed: https://gitlab.com/gitlab-org/gitlab/-/issues/290311

This module currently uses this endpoint to create a release in `publish.js`:

https://github.com/semantic-release/gitlab/blob/81cd6a88972e2e59ca6cf5d89ecea9d4529cdc9e/lib/publish.js#L71

This API call should be updated to use the [Release API](https://docs.gitlab.com/ee/api/releases/#create-a-release) instead.

nfriend:
### What does this PR do?

Resolves #183 by updating all requests to:

```
https://example.gitlab.com/api/v4/project/<project id>/repository/tags<tag name>/release`
```

to instead call:

```
https://example.gitlab.com/api/v4/project/<project id>/releases`
```

The former is deprecated and will be removed in GitLab 14.0 (in May 2021); see https://gitlab.com/gitlab-org/gitlab/-/issues/290311.

This has the added benefit of simplifying the process of uploading assets. The Releases API allows a release to be created with assets in a single request, removing the need to create assets separately.

This change is backwards compatible (the Releases API has been available since GitLab 11.7 - when the releases feature itself was delivered) and shouldn't impact current users of this plugin.

#### Testing

I've tested this change in a test project on GitLab.com:

- A release with assets: https://gitlab.com/nfriend/semantic-release-test-project/-/releases/v1.7.4
- A release without assets: https://gitlab.com/nfriend/semantic-release-test-project/-/releases/v1.8.0

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -67,23 +67,24 @@ module.exports = async (pluginConfig, context) => {
     );
   }
 
-  debug('Update git tag %o with commit %o and release description', gitTag, gitHead);
-  await got.post(urlJoin(gitlabApiUrl, `/projects/${encodedRepoId}/repository/tags/${encodedGitTag}/release`), {
+  debug('Create a release for git tag %o with commit %o', gitTag, gitHead);
+  await got.post(urlJoin(gitlabApiUrl, `/projects/${encodedRepoId}/releases`), {
     ...apiOptions,
-    json: {tag_name: gitTag, description: notes && notes.trim() ? notes : gitTag}, // eslint-disable-line camelcase
+    json: {
+      /* eslint-disable camelcase */
+      tag_name: gitTag,
+      description: notes && notes.trim() ? notes : gitTag,
+      assets: {
+        links: assetsList.map(({label, alt, url}) => {
+          return {
+            name: label || alt,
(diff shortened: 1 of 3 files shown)


nfriend:
@pvdlg Would you mind reviewing this PR? 🙏 

I'm not sure if `refactor:` is the correct commit message prefix for this kind of change 🤔. Happy to change it if something else makes sense.

Also cc: @travi and @gr2m in case @pvdlg is still away.

gr2m:
> I'm not sure if `refactor:` is the correct commit message prefix for this kind of change 🤔. Happy to change it if something else makes sense.

no worries, we can change the commit message with the squash & merge feature

gr2m (approved):
the PR looks good to me, tests look solid. Given that you work at GitLab and that you tested the changes I feel confident enough to go ahead and merge your PR.

Thank you for the great contribution!

nfriend:
Awesome, thanks @gr2m!

PR #185, opened 2020-12-09
fix: update release URL returned from publish.js

nfriend:
### Context

I noticed this while putting together https://github.com/semantic-release/gitlab/pull/184 and thought I'd open a separate issue to discuss.

### What does this PR do?

Updates the return value of `publish.js` to return a web URL to the _release_ instead of the _tag_. This is more consistent with `semantic-release/github`: https://github.com/semantic-release/github/blob/81847f1abd1c9714e28a17f712739ffebb586bec/lib/publish.js#L101

Here's an example of what a "tag" page looks like: https://gitlab.com/gitlab-org/gitlab/-/tags/v13.6.0-ee

And here's an example of its corresponding "release" page: https://gitlab.com/gitlab-org/gitlab/-/releases/v13.6.0-ee

The later is a nicer representation of the release, and we may be removing release notes from the former in the future: https://gitlab.com/gitlab-org/gitlab/-/issues/292279.

### Question

I looked through the plugin authoring docs, but I wasn't able to find how this return value is actually used. As a result, I'm not exactly sure what kind of change this is (`fix`, `feat`, or maybe even `BREAKING CHANGE`?).

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -88,5 +88,5 @@ module.exports = async (pluginConfig, context) => {
 
   logger.log('Published GitLab release: %s', gitTag);
 
-  return {url: urlJoin(gitlabUrl, encodedRepoId, `/tags/${encodedGitTag}`), name: 'GitLab release'};
+  return {url: urlJoin(gitlabUrl, encodedRepoId, `/-/releases/${encodedGitTag}`), name: 'GitLab release'};
 };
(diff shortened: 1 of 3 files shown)


nfriend:
@gr2m One more small PR for you, when you have time 🙂

gr2m:
> I looked through the plugin authoring docs, but I wasn't able to find how this return value is actually used. As a result, I'm not exactly sure what kind of change this is (`fix`, `feat`, or maybe even `BREAKING CHANGE`?).

Hmm, I don't think there is an obvious answer here, but I'd say it's a fix, because the intention was always to link to the release, and this change implements the intended behavior. Plus I don't think the change would be disruptive to any potential dependents

gr2m (approved):
Looks great 👍🏼

Issue #212, opened 2021-01-25
Cannot read property 'statusCode' of undefined


beselig:
## Current behavior

During verification process, when facing SSL problems, the following Error occurs:
```
TypeError: Cannot read property 'statusCode' of undefined
    at module.exports (/builds/....../test-ci-semantic-release/node_modules/@semantic-release/gitlab/lib/verify.js:65:26)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async verifyConditions (/builds/....../test-ci-semantic-release/node_modules/@semantic-release/gitlab/index.js:9:3)
    at async validator (/builds/....../test-ci-semantic-release/node_modules/semantic-release/lib/plugins/normalize.js:34:24)
    at async /builds/....../test-ci-semantic-release/node_modules/semantic-release/lib/plugins/pipeline.js:37:34
    at async Promise.all (index 0)
    at async next (/builds/....../test-ci-semantic-release/node_modules/p-reduce/index.js:16:18) {
  pluginName: '@semantic-release/gitlab'
}
```

## Expected behavior

The Error handling in `line 62` of [verify.js](https://github.com/semantic-release/gitlab/blob/master/lib/verify.js) is broken. It expects `error.response` to be always defined, which is not the case.

The catch block should not cause its own error, making the original error, which occured (i.e. due to our SSL issue), inaccessible.

Simple solution would be to just throw the error, when `error.response` is undefined. This is at least how we debugged this ourselves.

## Environment

- "@semantic-release/gitlab": "^6.0.8"
- **semantic-release** configuration:
```{
    "branches": ["master"],
    "plugins": [
        "@semantic-release/commit-analyzer",
        "@semantic-release/release-notes-generator",
        [
            "@semantic-release/gitlab",
            {
                "gitlabUrl": "https://gitlab.XYZ.net",
                "gitlabApiPathPrefix": "/api/v4"
            }
        ],
        "@semantic-release/npm",
        [
            "@semantic-release/git",
            {
                "assets": ["package.json"],
                "message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
            }
        ]
    ],
    "debug": true
}
```
- CI logs: 
```
TypeError: Cannot read property 'statusCode' of undefined
    at module.exports (/builds/....../test-ci-semantic-release/node_modules/@semantic-release/gitlab/lib/verify.js:65:26)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
[...]

Issue #186, opened 2020-12-11
 git ls-remote --heads https://gitlab-ci-token:... request fails


aldrashan:
I've followed the tutorial for Gitlab over at https://docs.gitlab.com/ee/ci/examples/semantic-release.html.
When running our pipeline, we're receiving an authentication error.
I've defined a GITLAB_TOKEN variable with the value set to my Project Access Token value.
I've defined a NPM_TOKEN variable with the value set to my Personal Access Token value.
I've defined a GITLAB_URL variable with the value set to 'https://gitlab.atsgroep.be/'.
I've defined a GITLAB_PREFIX variable with the value set to '/api/v4'.

Gitlab version: 13.3.6-ee (d7bc82f4b06)
Gitlab shell version: 13.6.0

> semantic-release --debug
[7:36:14 AM] [semantic-release] › ℹ  Running semantic-release version 17.3.0
2020-12-11T07:36:14.878Z semantic-release:config load config from: /builds/DEV/vanguard/.releaserc.json
2020-12-11T07:36:14.893Z semantic-release:config options values: { branches: [ 'master' ],
  repositoryUrl:
   'https://gitlab-ci-token:[secure]@gitlab.atsgroep.be/DEV/vanguard.git',
  tagFormat: 'v${version}',
  plugins:
   [ '@semantic-release/commit-analyzer',
     '@semantic-release/release-notes-generator',
     '@semantic-release/gitlab',
     '@semantic-release/npm',
     [ '@semantic-release/git', [Object] ] ],
  _: [],
  debug: true,
  '$0': 'node_modules/.bin/semantic-release' }
2020-12-11T07:36:15.386Z semantic-release:plugins options for @semantic-release/gitlab/verifyConditions: {}
[7:36:15 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/gitlab"
2020-12-11T07:36:15.392Z semantic-release:plugins options for @semantic-release/npm/verifyConditions: {}
[7:36:15 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/npm"
2020-12-11T07:36:15.393Z semantic-release:plugins options for @semantic-release/git/verifyConditions: { assets: [ 'package.json' ],
  message:
   'chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}' }
[...]

PR #214, opened 2021-01-27
fix: handle errors that don't have a `response`

nfriend:
### What does this PR do?

Currently, the error handling in `verify.js` assumes the error object will always contain a `response` property. This isn't always true; for example, SSL errors will throw an error object without a `response`.

This MR updates the error handling to check for `response` before accessing `response.statusCode`. If `response` doesn't exist, the error is re-thrown.

Related to https://github.com/semantic-release/gitlab/issues/186 and https://github.com/semantic-release/gitlab/issues/212

### DIFF
--- a/lib/verify.js
+++ b/lib/verify.js
@@ -59,9 +59,9 @@ module.exports = async (pluginConfig, context) => {
         errors.push(getError('EGLNOPERMISSION', {repoId}));
       }
     } catch (error) {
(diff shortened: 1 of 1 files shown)


nfriend:
@travi Would you mind reviewing this small PR? 🙏

travi (approved):
lgtm

PR #237, opened 2021-05-13
feat: support gitlab milestones in config

rjferguson21:
Adds support to attach milestones to the created GitLab Release via the plugin configuration

See: https://docs.gitlab.com/ee/api/releases/#create-a-release

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,12 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
 
 ### Options
 
-| Option                | Description                                                        | Default                                                                                                                                                                 |
-|-----------------------|--------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `gitlabUrl`           | The GitLab endpoint.                                               | `GL_URL` or `GITLAB_URL` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `https://gitlab.com`. |
-| `gitlabApiPathPrefix` | The GitLab API prefix.                                             | `GL_PREFIX` or `GITLAB_PREFIX` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `/api/v4`.      |
-| `assets`              | An array of files to upload to the release. See [assets](#assets). | -                                                                                                                                                                       |
+| Option                | Description                                                                                                                                    | Default                                                                                                                                                                 |
+|-----------------------|------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `gitlabUrl`           | The GitLab endpoint.                                                                                                                           | `GL_URL` or `GITLAB_URL` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `https://gitlab.com`. |
+| `gitlabApiPathPrefix` | The GitLab API prefix.                                                                                                                         | `GL_PREFIX` or `GITLAB_PREFIX` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `/api/v4`.      |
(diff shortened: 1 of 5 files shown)


nfriend:
@rjferguson21 This is really exciting! Sorry for the slow response here. I won't have time to take a look today or tomorrow, but I will prioritize this on Wednesday.

nfriend:
> but I will prioritize this on Wednesday.

So sorry - ran out of time today! But will get to this tomorrow or Friday!

nfriend:
@rjferguson21 Thanks for putting this together! 👏 

A couple of thoughts for you:

1. Let's update [the README](https://github.com/semantic-release/gitlab#options) with this new configuration option so that people know it's available.
2. I tested this out in a test project, and it works great! I'm curious how you plan on using this, though. Do you plan on hard-coding a specific milestone in your config? Or will you dynamically configure the `milestones` property in a `*.js` config? The reason I ask is that I'm having trouble thinking how I would automate this association. One pattern people use is to create a milestone with the same name as the release (for example, [this is what GitLab itself does](https://gitlab.com/gitlab-org/gitlab/-/releases)). But this would be difficult to do using this plugin, since the release version is managed by `semantic-release` 🤔 This isn't a reason _not_ to add this ability, but I just want to make sure I understand what kinds of workflows this will enable.

nfriend (changes requested):
@rjferguson21 Thanks for putting this together! 👏 

A couple of thoughts for you:

1. Let's update [the README](https://github.com/semantic-release/gitlab#options) with this new configuration option so that people know it's available.
2. I tested this out in a test project, and it works great! I'm curious how you plan on using this, though. Do you plan on hard-coding a specific milestone in your config? Or will you dynamically configure the `milestones` property in a `*.js` config? The reason I ask is that I'm having trouble thinking how I would automate this association. One pattern people use is to create a milestone with the same name as the release (for example, [this is what GitLab itself does](https://gitlab.com/gitlab-org/gitlab/-/releases)). But this would be difficult to do using this plugin, since the release version is managed by `semantic-release` 🤔 This isn't a reason _not_ to add this ability, but I just want to make sure I understand what kinds of workflows this will enable.

rjferguson21:
Thanks for looking at this and the feedback @nfriend.  I've added the `milestones` option to the README. 

Your second question is valid, we're using a job like this:
```
release:
  image: node:lts-alpine3.13
  stage: release
  script:
    - apk add jq moreutils curl git
    - |
      export MILESTONE=$(curl --header "PRIVATE-TOKEN: $RELEASE_TOKEN" $GITLAB_URL/api/v4/groups/$GROUP_ID/milestones\?state\=active | jq .[-1].title)
    - jq ".plugins[4][1].milestones = [$MILESTONE]" .releaserc.json | sponge .releaserc.json
    - npm install -g semantic-release @semantic-release/git https://github.com/rjferguson21/gitlab#add-milestones @semantic-release/commit-analyzer @semantic-release/release-notes-generator @semantic-release/changelog
    - semantic-release
  only:
    - master
```

The idea is to associate project releases to a group level milestone, so this script block tries to go fetch the current active milestone and attaches the particular project's release to a group level milestone.  It is a bit challenging because semantic-release does not support passing in arguments to plugins via the command line, so I am having to mutate the project's release config in CI (I chose the `.json` version).

nfriend (approved):

nfriend:
@rjferguson21 Awesome, thanks for updating the documentation!

And thanks for the details about automating the associating to the active group level milestone.

This PR looks good to me 👍 

@travi I noticed I now have permission to merge in this repo. Is there any particular process I should be following before merging? Or would you like me to continue pinging one of you for final review/merge?

nfriend:
cc @pvdlg @gr2m in case you can answer my [question about the review process above](https://github.com/semantic-release/gitlab/pull/237#issuecomment-846295696) 😄

travi:
@nfriend sorry for missing your question before. you are welcome to merge if you believe that contributions are ready for release. ideally, the test suite should protect us from breakage. if we ever find that to be untrue, the goal would be to close the holes that we find. publishing a revert or a fix is an option in the cases where something slips through. 

you also now have triage rights across all semantic-release repos. we appreciate the help you've been giving for this plugin. up to you if you want to participate in other areas.

nfriend:
Awesome, thanks @travi!

And thanks for this @rjferguson21, merging now 👍

PR #240, opened 2021-05-26
fix(publish): Compatible with older versions of GitLab, lack of "name" attribu…

ldc2726:
Compatible with older versions of GitLab, lack of "name" attribute, request result return 400 codete, request result return 400 code

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -14,7 +14,7 @@ module.exports = async (pluginConfig, context) => {
   const {
     cwd,
     options: {repositoryUrl},
-    nextRelease: {gitTag, gitHead, notes},
+    nextRelease: {gitTag, name, gitHead, notes},
     logger,
   } = context;
   const {gitlabToken, gitlabUrl, gitlabApiUrl, assets} = resolveConfig(pluginConfig, context);
@@ -25,7 +25,8 @@ module.exports = async (pluginConfig, context) => {
   const apiOptions = {headers: {'PRIVATE-TOKEN': gitlabToken}};
 
   debug('repoId: %o', repoId);
-  debug('release name: %o', gitTag);
+  debug('release name: %o', name)
+  debug('release gitTag: %o', gitTag);
   debug('release ref: %o', gitHead);
 
   if (assets && assets.length > 0) {
@@ -72,6 +73,7 @@ module.exports = async (pluginConfig, context) => {
     ...apiOptions,
     json: {
       /* eslint-disable camelcase */
+      name:name ? name : gitTag,
       tag_name: gitTag,
       description: notes && notes.trim() ? notes : gitTag,
       assets: {


lib/publish.js
```diff
@@ -72,6 +73,7 @@ module.exports = async (pluginConfig, context) => {
     ...apiOptions,
     json: {
       /* eslint-disable camelcase */
+      name:name ? name : gitTag,
```
nfriend, line 76: Thanks for this - this matches GitLab's default behavior in version 12+.

**nitpick:** I personally like this phrasing better:

```suggestion
      name: name || gitTag
```

lib/publish.js
```diff
@@ -25,7 +25,8 @@ module.exports = async (pluginConfig, context) => {
   const apiOptions = {headers: {'PRIVATE-TOKEN': gitlabToken}};
 
   debug('repoId: %o', repoId);
-  debug('release name: %o', gitTag);
+  debug('release name: %o', name)
```
nfriend, line 28: **small issue:** if `name` is falsy, this will output something like:

```
release name: undefined
```

However, the actual network request will be made with `name: gitTag`.

Let's make this log message consistent with what's actually being sent by computing the name above this line and then using the computed name here and in the actual network request:

```js
const releaseName = name || gitTag;

// ...

debug('release name: %o', releaseName)

// ...

json: {
  name: releaseName
}
```

nfriend (changes requested):
Thanks for this @ldc2726, looking good!

Two small comments in `publish.js`. It also looks like there are a few PR checks failing - can you take a look?

Also, let's make sure this new logic has test coverage in both `publish.test.js` and `integration.test.js`.

Back to you for now!

nfriend:
@ldc2726 Is this ready for another review? Looks like the checks are still failing so I'm not sure if I should take another look. Feel free to ping me when this is ready 😄

alecgerona:
Maybe we can also log the failure reason?

nfriend:
> Maybe we can also log the failure reason?

@alecgerona I opened a PR to do some better logging in https://github.com/semantic-release/gitlab/pull/245 👍

nfriend:
@ldc2726 Let me know if you need any help getting the merge checks green.

nfriend:
@ldc2726 Sorry for the slow turnaround time on this PR - I've been out for the last several weeks.

After some discussion about this change (see https://github.com/semantic-release/gitlab/pull/246), we've decided not to merge this fix since it's only applicable to an officially unsupported version of GitLab (v11). This project's README has since been updated to clarify our [compatibility guidelines](https://github.com/semantic-release/gitlab/#compatibility):

> The latest version of this plugin is compatible with all currently-supported versions of GitLab, [which is the current major version and previous two major versions](https://about.gitlab.com/support/statement-of-support.html#version-support). This plugin is not guaranteed to work with unsupported versions of GitLab.

This guideline allows us to minimize the maintenance burden of this plugin, since all of the maintainers of this plugin are volunteers.

Sorry for the inconvenience this is causing/will cause you! I appreciate your participation in this project.

Issue #167, opened 2020-11-05
Allow custom names for assets


leoshusar:
Hello, I've seen that `semantic-release/github` has more options for assets, it has `name` property and both properties `name` and `label` are allowed to insert next version by for example `${nextRelease.gitTag}`.

Is something like this planned for this plugin too? 

It would be handy for me because I could've keep my binary archive named like `build.tar.gz` and then rename it to `MyApp-v1.0.0.tar.gz` by this plugin - mainly because `path` property cannot be dynamic even in `github` plugin.

Thank you!

nfriend:
@leoshusar I think this would be a great feature to have 👍 It would be nice to maintain parity with `@semantic-release/github` in this case.

Would you be open to putting together a PR for this?

Issue #161, opened 2020-08-31
Supporting assets release types


MathieuPuech:
It has been introduced in GitLab 13.1, we can now specify the asset type of a link.

See https://docs.gitlab.com/ee/user/project/releases/#links

nfriend:
@MathieuPuech Agreed, this would be a great feature to have! (And I think it would be relatively easy to implement.)

Would you be interested in opening up a PR for this?

PR #242, opened 2021-06-01
feat: support asset type and filepath

yxonic:
Implemented two options for assets according to GitLab [document on asset permanent link and link type](https://git.bdaa.pro/help/user/project/releases/index#permanent-links-to-release-assets), addressing #161.

It can also be a solution to #167, with which now we can download asset using URL like this: `https://host/namespace/project/releases/v1.0.0/downloads/build.tar.gz`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` 
 | -------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
 | `path`   | **Required.** A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
 | `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file.| File name extracted from the `path`. |
+| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://git.bdaa.pro/help/user/project/releases/index#release-assets)). | `other` |
+| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://git.bdaa.pro/help/user/project/releases/index#permanent-links-to-release-assets)). | - |
 
 Each entry in the `assets` `Array` is globbed individually. A [glob](https://github.com/isaacs/node-glob#glob-primer)
 can be a `String` (`"dist/**/*.js"` or `"dist/mylib.js"`) or an `Array` of `String`s that will be globbed together
--- a/README.md
+++ b/README.md
@@ -74,8 +74,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` 
 | -------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
 | `path`   | **Required.** A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
 | `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file.| File name extracted from the `path`. |
-| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://git.bdaa.pro/help/user/project/releases/index#release-assets)). | `other` |
-| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://git.bdaa.pro/help/user/project/releases/index#permanent-links-to-release-assets)). | - |
+| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). | `other` |
+| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. | - |
 
 Each entry in the `assets` `Array` is globbed individually. A [glob](https://github.com/isaacs/node-glob#glob-primer)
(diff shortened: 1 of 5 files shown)


lib/publish.js
```diff
@@ -77,10 +77,12 @@ module.exports = async (pluginConfig, context) => {
       description: notes && notes.trim() ? notes : gitTag,
       milestones,
       assets: {
-        links: assetsList.map(({label, alt, url}) => {
+        links: assetsList.map(({label, alt, url, type, filepath}) => {
           return {
             name: label || alt,
             url: urlJoin(gitlabUrl, repoId, url),
+            link_type: type || 'other',
```
nfriend, line 84: I can't remember off the top of my head if old versions of GitLab (i.e. before `link_type` was introduced) will accept POST requests that include fields it doesn't recognize 🤔 

If they _don't_, this will cause an error for older version of GitLab. But I'm about 90% sure `link_type` would just be ignored.

Regardless, I think it would be best to just pass `type` here:

```suggestion
            link_type: type
```

This has two advantages:

1. It defers to GitLab to determine what the default should be if no `link_type` is passed. No need to update this code if the default changes on GitLab's side
2. It passes `undefined` to the API if no `link_type` is configured, side stepping the issue mentioned above (I think - I haven't tested this)

What do you think?

README.md
```diff
@@ -74,6 +74,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
 | -------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
 | `path`   | **Required.** A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
 | `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file.| File name extracted from the `path`. |
+| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://git.bdaa.pro/help/user/project/releases/index#release-assets)). | `other` |
```
nfriend, line 77: Awesome, thanks for updating the docs!

nfriend (changes requested):
@yxonic This looks great! Thanks for putting this PR together.

I have one small suggestion on the current changes. I also think we should cover this new field in a test, most likely in `publish.test.js`.

Back to you for now 🏓 Feel free to ping me when this is ready for another review!

yxonic:
@nfriend thanks for your time reviewing the code! I've made several changes according to your suggestions:

1. I agree with your comments, so `link_type` is now undefined as default, which also fixes tests.
2. Added a testcase to cover these fields. I'm not exactly sure if I did it right, so if there's any problem, just let me know :)

lib/publish.js
```diff
@@ -77,10 +77,12 @@ module.exports = async (pluginConfig, context) => {
       description: notes && notes.trim() ? notes : gitTag,
       milestones,
       assets: {
-        links: assetsList.map(({label, alt, url}) => {
+        links: assetsList.map(({label, alt, url, type, filepath}) => {
           return {
             name: label || alt,
             url: urlJoin(gitlabUrl, repoId, url),
+            link_type: type,
+            filepath,
```
nfriend, line 85: One small issue I discovered while testing this branch in a test project. If `.releaserc.json` defines its assets using a wildcard in the `path`:

```json
"assets": [
  {
    "path": "images/**/*",
    "type": "image",
    "filepath": "/path/to/image"
  }
]
```

and if the wildcard matches more than one file, GitLab will throw an error when creating the release. This is because behind the scenes, this plugin creates a separate asset link entry for each matched file, and each entry will have the same `type` and `filepath`. However, Gitlab requires that the `filepath` be unique in the context of a release.

Unfortunately this causes kind of a generic error message when running `semantic-release` - it wouldn't be very clear to users what's going wrong.

So, I think we should:

1. Document this limitation
2. Validate against this and throw a helpful error message so that it's obvious to end users what's going wrong and how to fix it

I think the easiest way to validate this would be to add a validator to `verify.js` that disallows the `filepath` keyword for asset definitions with a `*` in the `path`.

Another option might be to throw an error in `publish.js` after we've expanded the `path` pattern. This would allow the `filepath` option to succeed provided the `path` only resolves to a single file. But I think this might be a source for frustration - in the future, the path may start to match more than one file, which would cause the pipeline to break. So I think the `verify.js` option is our best bet.

Are you up for taking this on? Let me know if you need help.

nfriend (changes requested):
@yxonic Looking great! Thanks for the changes and for adding the test.

I did find one usability issue that I think we should address, which I've mentioned below.

Also, I think we should log the `type` and `filepath` when in `debug` mode on lines 54-55 in `publish.js` (for some reason GitHub doesn't let me comment on the lines directly - I guess I'm just too used to GitLab 😅).

Let me know what you think. And please let me know if you need help with any of the suggested changes. I realize I'm probably requesting more changes than you originally bargained for 😄

README.md
```diff
@@ -74,6 +74,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
...
+| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://git.bdaa.pro/help/user/project/releases/index#permanent-links-to-release-assets)). | - |
```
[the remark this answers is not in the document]
yxonic: sorry for the wrong links😅 i'll update them

lib/publish.js
yxonic: I can spare time to fix this in 2 days or so. I will also include some more tests.

yxonic:
@nfriend Hi~ I just fixed the links. regarding the duplicate `filepath` issue, I took a slightly different approach here. I noticed that `label` is basically ignored when there is more than one files, and I think we can do the same for `filepath` for consistency: `filepath` will still work when there is only one file matched (user can still use `*`), and won't break the pipeline if more than one file is matched later. This behavior is also documented and tested.

nfriend:
@yxonic Sorry for the slow response here, I'm going to try and get to this tomorrow!

nfriend:
> Regarding the duplicate filepath issue, I took a slightly different approach here.

@yxonic Good call! This is a simpler and more consistent solution 👍 And thanks for documenting this behavior.

nfriend (approved):
@yxonic This looks good to me! Just gave this a try in a test project and it works great. Thanks for all your work here!

Issue #239, opened 2021-05-20
code 400


ldc2726:
021-05-20T05:02:43.750Z semantic-release:git commited files: [ 'CHANGELOG.md' ]
[1:02:44 PM] [semantic-release] [@semantic-release/git] › ℹ  Prepared Git release: v1.2.5
[1:02:44 PM] [semantic-release] › ✔  Completed step "prepare" of plugin "@semantic-release/git"
[1:02:44 PM] [semantic-release] › ℹ  Start step "generateNotes" of plugin "@semantic-release/release-notes-generator"
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator version: '1.2.5'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator host: undefined
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator owner: 'dyWeb'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator repository: 'DataCockpit/demo'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator previousTag: 'v1.2.4'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator currentTag: 'v1.2.5'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator host: 'http://192.168.1.184'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator linkReferences: undefined
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator issue: 'issues'
2021-05-20T05:02:44.648Z semantic-release:release-notes-generator commit: 'commit'
[1:02:44 PM] [semantic-release] › ✔  Completed step "generateNotes" of plugin "@semantic-release/release-notes-generator"
[1:02:46 PM] [semantic-release] › ✔  Created tag v1.2.5
[1:02:46 PM] [semantic-release] › ℹ  Start step "publish" of plugin "@semantic-release/gitlab"
2021-05-20T05:02:46.971Z semantic-release:gitlab repoId: 'dyWeb/DataCockpit/demo'
2021-05-20T05:02:46.971Z semantic-release:gitlab release name: 'v1.2.5'
2021-05-20T05:02:46.971Z semantic-release:gitlab release ref: '852ee054f6c68bab2084651f7598d2f74ad2face'
2021-05-20T05:02:46.971Z semantic-release:gitlab Create a release for git tag 'v1.2.5' with commit '852ee054f6c68bab2084651f7598d2f74ad2face'
[...]

PR #246, opened 2021-06-17
docs: add note about version compatibility

nfriend:
### What does this PR do?

Adds a note about which versions of GitLab are supported by this plugin, and also adds a note about a specific breaking change in GitLab 14.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,4 +110,4 @@ If you are using GitLab.com or have upgraded your self-hosted GitLab instance to
 
 #### Why?
 
-In GitLab 14.0, creating a release using the [Tags API](https://docs.gitlab.com/ee/api/tags.html) has been removed (see [<span>#</span>290311](https://gitlab.com/gitlab-org/gitlab/-/issues/290311)). This plugin was updated to use the Release API instead in https://github.com/semantic-release/gitlab/pull/184, which is available in version `6.0.7` and beyond.
+In GitLab 14.0, creating a release using the [Tags API](https://docs.gitlab.com/ee/api/tags.html) has been removed (see [<span>#</span>290311](https://gitlab.com/gitlab-org/gitlab/-/issues/290311)). This plugin was updated to use the [Releases API](https://docs.gitlab.com/ee/api/releases/#create-a-release) instead in https://github.com/semantic-release/gitlab/pull/184, which is available in version `6.0.7` and beyond.
--- a/README.md
+++ b/README.md
@@ -99,3 +99,15 @@ distribution` and `MyLibrary CSS distribution` in the GitLab release.
 `[['dist/**/*.{js,css}', '!**/*.min.*'], {path: 'build/MyLibrary.zip', label: 'MyLibrary'}]`: include all the `js` and
 `css` files in the `dist` directory and its sub-directories excluding the minified version, plus the
 `build/MyLibrary.zip` file and label it `MyLibrary` in the GitLab release.
+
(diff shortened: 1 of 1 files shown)


nfriend:
@keplersj @travi Can you please review this proposed compatibility guideline? 🙏 

For context, some of the users of this plugin are seeing errors due to using an unsupported version of GitLab: https://github.com/semantic-release/gitlab/issues/239.

Let me know if this seems reasonable (and consistent with other `semantic-release` plugins).

travi:
this looks good, although i am a little confused by the link to the issue. it sounds like you are referring to the later comment specifically about the v14 change. there is also a mention of v11 support in that thread and this seems to contradict with entertaining the PR to continue enabling v11 support beyond EOL. i'm supportive of this language and dropping support of gitlab versions that are EOL, but just want to make sure that aligns with what we decide to do with that PR. do we intend to close that PR as part of this?

travi (approved):
i like this messaging, but lets resolve the question about v11 before merging

nfriend:
> here is also a mention of v11 support in that thread and this seems to contradict with entertaining the PR to continue enabling v11 support beyond EOL.

@travi Sorry, yes, that is a little confusing - I should have explained in more detail. Since https://github.com/semantic-release/gitlab/pull/240 was already in motion, and since it's a forwards-compatible change, I thought we could make an exception (especially since support for v11 _just_ ended) and get this fix merged. But going forward, we wouldn't entertain any v11-specific fixes.

The v14 breaking change is a separate issue, but I thought I'd address it in the same section since they're both related to the version of the GitLab instance.

travi:
> I thought we could make an exception (especially since support for v11 _just_ ended) and get this fix merged. But going forward, we wouldn't entertain any v11-specific fixes.

i think that is fair if that ends up working out. it seems low maintenance if it does merge, but i think it would also be reasonable if it didnt merge. you can make the call on that piece.

whether dropping support for gitlab versions constitutes a breaking change or not is something we should also define sometime. i lean toward yes, but that would mean taking steps at the appropriate times to publish the breaking change. it may make sense to hold off until a change is made that is actually incompatible with an older version, but i wouldnt want to carry the burden of testing against end-of-life versions to know when that happens. thats the balance that makes me hesitate slightly about merging something to support a version that is already (even if only recently) EOL, since it becomes questionable when it is safe to remove vs keeping it around forever for the sake of being uncertain.

all of that is not to convince you to go one way or another, but rather just to share some of the thoughts on my mind around the topic.

nfriend:
@travi Thanks for your thoughts!

> whether dropping support for gitlab versions constitutes a breaking change or not is something we should also define sometime. i lean toward yes, but that would mean taking steps at the appropriate times to publish the breaking change.

I agree, I think a breaking change would make the most sense in that scenario.

>  thats the balance that makes me hesitate slightly about merging something to support a version that is already (even if only recently) EOL, since it becomes questionable when it is safe to remove vs keeping it around forever for the sake of being uncertain.

Thanks for this input! That makes a lot of sense. I'm now leaning toward closing https://github.com/semantic-release/gitlab/pull/240 to avoid the maintenance burden you mentioned above.

---

Is there anything else you'd like to see outlined in this docs update? Or are you okay if I merge this in its current state?

travi:
> Is there anything else you'd like to see outlined in this docs update? Or are you okay if I merge this in its current state?

this seems great for this stage. thanks for putting this together!

Issue #192, opened 2020-12-17
Error in request to create release


tacnoman:
Hello everyone

I'm receiving an error when I run the semantic release in last step. The code can create the tag, but after this, there is a request to `/release` that I'm receiving a 400 error code

```
[3:27:32 PM] [semantic-release] [@semantic-release/changelog] › ℹ  Create /tmp/build/group/project-name/CHANGELOG.md
[3:27:32 PM] [semantic-release] › ✔  Completed step "prepare" of plugin "@semantic-release/changelog"
[3:27:33 PM] [semantic-release] › ✔  Created tag v1.3.1
[3:27:33 PM] [semantic-release] › ℹ  Start step "publish" of plugin "@semantic-release/gitlab"
[3:27:33 PM] [semantic-release] › ✖  Failed step "publish" of plugin "@semantic-release/gitlab"
[3:27:33 PM] [semantic-release] › ✖  An error occurred while running semantic-release: HTTPError: Response code 400 (Bad Request)
    at EventEmitter.<anonymous> (/usr/local/lib/node_modules/@semantic-release/gitlab/node_modules/got/dist/source/as-promise.js:118:31)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  pluginName: '@semantic-release/gitlab'
}
HTTPError: Response code 400 (Bad Request)
    at EventEmitter.<anonymous> (/usr/local/lib/node_modules/@semantic-release/gitlab/node_modules/got/dist/source/as-promise.js:118:31)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  pluginName: '@semantic-release/gitlab'
}
```

I start to investigate the problem in code and I put a debug with error message in API:

```
2020-12-17T16:59:53.675Z semantic-release:gitlab error in request: Response code 400 (Bad Request)
2020-12-17T16:59:53.675Z semantic-release:gitlab {"error":"name is missing"}
```

Is required a field called `name` in json data send with GOT.

https://github.com/semantic-release/gitlab/blob/f8957c30a29f851e9400a6a3bbe169d9b9c7cfbb/lib/publish.js#L75
There is no field called `name` in this file

I changed and add a name with any value and I received a 201 code.

Is this a bug, different version or I made something wrong?

Thanks for this

nfriend:
Thanks for this detailed report @tacnoman!

[...]

PR #261, opened 2021-07-23
fix(pipeline): Add name parameter to avoid 400 error

renatocassino:
This PR is related with this issue.

https://github.com/semantic-release/gitlab/issues/192

The Gitlab api is returning error when make a request. The status code is 400 and the error is:

```json
{"error":"name is missing"}
```

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -88,6 +88,7 @@ module.exports = async (pluginConfig, context) => {
   const json = {
     /* eslint-disable camelcase */
     tag_name: gitTag,
+    name: `Bump tag: ${gitTag}`,
     description: notes && notes.trim() ? notes : gitTag,
     milestones,
     assets: {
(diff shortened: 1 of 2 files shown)


merlosy:
This plugin doesn' work anymore without the "name" for all newer version of Gitlab.

lib/publish.js
```diff
@@ -88,6 +88,7 @@ module.exports = async (pluginConfig, context) => {
   const json = {
     /* eslint-disable camelcase */
     tag_name: gitTag,
+    name: `Bump tag: ${gitTag}`,
```
merlosy, line 91: The name could **optionally** be resolved by the plugin config so that:
- if it is defined, the given value is set
- if it is not defined, this value is assigned by default
like:
```js
name: pluginConfig.name || `Bump tag: ${gitTag}`,
```

Issue #265, opened 2021-07-27
Can not create release if ** matches multiple files with the same name.


EvanCarroll:
I'm getting this error,

```text
[12:29:11 PM] [semantic-release] [@semantic-release/gitlab] › ✖  An error occurred while making a request to the GitLab release API:
HTTPError: Response code 400 (Bad Request)
    at EventEmitter.<anonymous> (/home/ecarroll/n/lib/node_modules/@semantic-release/gitlab/node_modules/got/dist/source/as-promise.js:118:31)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
[12:29:11 PM] [semantic-release] › ✖  Failed step "publish" of plugin "@semantic-release/gitlab"
[12:29:11 PM] [semantic-release] › ✖  An error occurred while running semantic-release: HTTPError: Response code 400 (Bad Request)
    at EventEmitter.<anonymous> (/home/ecarroll/n/lib/node_modules/@semantic-release/gitlab/node_modules/got/dist/source/as-promise.js:118:31)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  pluginName: '@semantic-release/gitlab'
}
HTTPError: Response code 400 (Bad Request)
    at EventEmitter.<anonymous> (/home/ecarroll/n/lib/node_modules/@semantic-release/gitlab/node_modules/got/dist/source/as-promise.js:118:31)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  pluginName: '@semantic-release/gitlab'
}%
```_

This error comes right after a release `PUSH`. Currently it's **silently failing** failing I opened this ticket to fix the silent failing: https://github.com/semantic-release/gitlab/issues/264

```json
{
  responseType: 'text',
  body: '{"message":"Validation failed: Links have duplicate values (acme-angular-validators.js) for #\\u003cRelease:0x00007f0702a43908\\u003e scope"}'
}```

You can see here that the problem is that GitLab errors with `400 (Bad Request)` because it's upload multiple files (in different directories) to the release API.

```text
dist/acme/angular-validators/fesm2015/acme-angular-validators.js
dist/acme/angular-validators/esm2015/acme-angular-validators.js
```

EvanCarroll:
This is the same issue as https://github.com/semantic-release/gitlab/issues/158

Note:
[...]

Issue #158, opened 2020-08-06
An error occurred while running semantic-release: HTTPError: Response code 400 (Bad Request)


digitaltopo:
Seeing a `HTTPError: Response code 400 (Bad Request)` error in the GitLab pipeline job:

```bash
 [9:56:02 PM] [semantic-release] › ✖  Failed step "publish" of plugin "@semantic-release/gitlab"
 [9:56:02 PM] [semantic-release] › ✖  An error occurred while running semantic-release: HTTPError: Response code 400 (Bad Request)
     at EventEmitter.<anonymous> (/builds/code/path/to/lib/node_modules/got/dist/source/as-promise.js:118:31)
     at processTicksAndRejections (internal/process/task_queues.js:93:5) {
   pluginName: '@semantic-release/gitlab'
 }
 HTTPError: Response code 400 (Bad Request)
     at EventEmitter.<anonymous> (/builds/code/path/to/lib/node_modules/got/dist/source/as-promise.js:118:31)
     at processTicksAndRejections (internal/process/task_queues.js:93:5) {
   pluginName: '@semantic-release/gitlab'
 }
 ERROR: Job failed: exit code 1
```

I belive error is triggered from `publish.js` file in `semantic-release/gitlab`:
https://github.com/semantic-release/gitlab/blob/8477d146c47d9c515ae12bb98949fc535a84435d/lib/publish.js

Project's `.releaserc` looks like this:

```json
{
  "repositoryUrl": ...,
  "branches": [
    "master",
    "next"
  ],
  "prepare": [
    "@semantic-release/changelog",
    "@semantic-release/npm",
    {
      "path": "@semantic-release/git",
      "assets": [
        "package.json",
        "package-lock.json",
        "CHANGELOG.md"
      ],
      "message": "chore(release): ${nextRelease.version} [skip ci]nn${nextRelease.notes}"
    }
  ],
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    "@semantic-release/changelog",
    [
      "@semantic-release/npm",
      {
        "npmPublish": true,
        "pkgRoot": "./lib"
      }
    ],
    [
      "@semantic-release/git",
      {
        "assets": [
          "package.json",
          "CHANGELOG.md"
        ],
        "message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
      }
    ],
    [
      "@semantic-release/gitlab",
      {
        "assets": [
          {
[...]

PR #267, opened 2021-07-29
fix: file labels in GitLab releases, to ensure they're unique.

EvanCarroll:
Previously GitLab lables were just the basename for files uploaded as
part of the release. This is problematic because GitLab doesn't allow
conflicting labels -- a condition that could be caused by uploading a
release with two files by the same name in different directories. This
would generate a 409 Conflict error.

This changes the labels for files uploaded as part of a release to the
name relative to pkgRoot, or the package.

A project may look like this

  pkg
  pkg \ foo \ baz
  pkg \ bar \ baz

This would previously result in two conflicting labels of 'baz'. Now you
would have {"foo/baz", "bar/baz"} with no conflict.

GitHub issues: #265, #158

### DIFF
--- a/lib/glob-assets.js
+++ b/lib/glob-assets.js
@@ -1,12 +1,11 @@
 const path = require('path');
-const {basename} = require('path');
 const {isPlainObject, castArray, uniqWith, uniq} = require('lodash');
 const dirGlob = require('dir-glob');
 const globby = require('globby');
 const debug = require('debug')('semantic-release:gitlab');
 
-module.exports = async ({cwd}, assets) =>
-  uniqWith(
+module.exports = async ({pkgRoot, cwd}, assets) => {
+  return uniqWith(
     []
       .concat(
         ...(await Promise.all(
@@ -42,7 +41,7 @@ module.exports = async ({cwd}, assets) =>
                 // - `filepath` ignored (also to avoid duplicates)
                 // - other properties of the original asset definition
                 const {filepath, ...others} = asset;
-                return globbed.map(file => ({...others, path: file, label: basename(file)}));
+                return globbed.map(file => ({...others, path: file, label: path.relative(pkgRoot || '.', file)}));
               }
 
               // If asset is an Object, output an Object definition with:
@@ -66,3 +65,4 @@ module.exports = async ({cwd}, assets) =>
     // Compare `path` property if Object definition, value itself if String
     (a, b) => path.resolve(cwd, isPlainObject(a) ? a.path : a) === path.resolve(cwd, isPlainObject(b) ? b.path : b)
   );
+};
--- a/lib/resolve-config.js
+++ b/lib/resolve-config.js
@@ -1,23 +1,21 @@
 const {castArray, isNil} = require('lodash');
 const urlJoin = require('url-join');
 
-module.exports = (
-  {gitlabUrl, gitlabApiPathPrefix, assets, milestones},
-  {
-    envCi: {service} = {},
-    env: {
-      CI_PROJECT_URL,
-      CI_PROJECT_PATH,
-      CI_API_V4_URL,
-      GL_TOKEN,
-      GITLAB_TOKEN,
-      GL_URL,
-      GITLAB_URL,
-      GL_PREFIX,
-      GITLAB_PREFIX,
-    },
-  }
-) => {
+module.exports = (pluginConfig, context) => {
+  const {gitlabUrl, gitlabApiPathPrefix, assets, milestones, pkgRoot} = pluginConfig;
+  const {service} = context.envCi || {service: undefined};
+  const {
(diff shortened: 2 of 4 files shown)


EvanCarroll:
The force pushes are all commit message. I linked it to the issues and fixed some typos.

EvanCarroll:
It may be worth documenting too, but you would use this feature like this

```text
      "gitlabUrl": "https://gitlab.awe.eco.cpanel.net/",
      "pkgRoot": "dist/cpanel/angular-validators",
      "assets": [
        {"path": "dist/cpanel/angular-validators/**", "label": "Dist output"}
      ]   
    }] 
```

Now all things found in `dist/cpanel/angular-validators/` have their label in the release relative to the `pkgRoot`

EvanCarroll:
I rebased this can I get some input on it?

fgreinacher:
This PR has been stale for nearly two years now, so I'm closing it.

Feel free to send a new PR if you want this 🙇

EvanCarroll:
It was stale because you never gave any input on it lol.

fgreinacher:
Ha wow, too ambitious in cleaning up. Sorry!

lib/glob-assets.js
```diff
@@ -42,7 +41,7 @@ module.exports = async ({cwd}, assets) =>
                 // - `filepath` ignored (also to avoid duplicates)
                 // - other properties of the original asset definition
                 const {filepath, ...others} = asset;
-                return globbed.map(file => ({...others, path: file, label: basename(file)}));
+                return globbed.map(file => ({...others, path: file, label: path.relative(pkgRoot || '.', file)}));
```
fgreinacher, line 44: Should we consider this a breaking change?

lib/resolve-config.js
```diff
@@ -31,6 +29,7 @@ module.exports = (
       : 'https://gitlab.com');
 
   return {
+    pkgRoot,
```
fgreinacher, line 32: I understand you are mirroring the NPM plugin here, but it feels a bit displaced here. The concept of a package does not really exist for the GitLab plugin.
What about `assetBasePath` or `assetRootPath`?

test/resolve-config.test.js
```diff
@@ -224,6 +234,7 @@ test('Ignore GitLab CI/CD environment variables if not running on GitLab CI/CD',
       gitlabApiUrl: urlJoin('https://gitlab.com', '/api/v4'),
       assets: undefined,
       milestones: undefined,
+      pkgRoot: undefined,
     }
   );
 });
```
fgreinacher, line 240: Can you add some test cases for the new option? Especially weird cases like when the asset is not within the base/root path or when the base/root path is somehow malformed/does not exist.

lib/glob-assets.js
```diff
@@ -1,12 +1,11 @@
...
-const {basename} = require('path');
 const {isPlainObject, castArray, uniqWith, uniq} = require('lodash');
 const dirGlob = require('dir-glob');
 const globby = require('globby');
 const debug = require('debug')('semantic-release:gitlab');
 
-module.exports = async ({cwd}, assets) =>
-  uniqWith(
+module.exports = async ({pkgRoot, cwd}, assets) => {
+  return uniqWith(
```
fgreinacher, line 8: This can stay an arrow function, no?

lib/resolve-config.js
fgreinacher: Also we some documentation please 🙇

fgreinacher (changes requested):
Thanks @EvanCarroll and sorry that I somewhat missed that this is ready for review.

I left some comments :)

lib/glob-assets.js
EvanCarroll: @fgreinacher I'm confused. It's still an arrow function

lib/glob-assets.js
EvanCarroll: @fgreinacher I wouldn't unless generating duplicate labels was support somewhere and things depended o nit. Honestly don't know. Seems weird. Imagine creating a listing of files in multiple directories and losing the directory they were in, that was the old behavior.  So the labels would collide and the SCM would reject the operation.

JonasSchubert:
@EvanCarroll shall somebody else take over for you or do you want to continue here?

fgreinacher:
I think somebody else needs to take over here.

EvanCarroll:
Please, take it over. I'm not adding tests for this. If you want to add tests to a project that didn't previously have them, that's another issue imho.

Besides I have to maintain a forks of all of the bugs I've fixed in the semantic-release projects anyway.

lib/resolve-config.js
bever1337: There's no documentation anywhere else in this module. That's a separate issue.

PR #279, opened 2021-09-22
feat(node-version): raised the minimum required version to v14.17

greg0ire:


languitar:
This would then fix #280

travi:
Thank you for the contribution. There are a few additional changes that we need to include along with these. Would you mind taking a look at https://github.com/semantic-release/github/pull/410/files as an example and including the engines related changes as well?

package.json
```diff
@@ -70,7 +70,7 @@
     "all": true
   },
   "peerDependencies": {
-    "semantic-release": ">=15.8.0 <18.0.0"
+    "semantic-release": ">=15.8.0 <19.0.0"
```
travi, line 73: because of the `engines` change above, only v18 and above satisfy the needs of this package. 
```suggestion
    "semantic-release": ">=18.0.0"
```

travi:
thanks for the update of the engines definition. could you also include the changes to the workflow file so that we run verification against supported node versions, ensuring that our dependency tree continues to meet the updated engines requirements?

> Do I need to also touch the lock file?

yes, the lockfile will need to be updated because of the update to the installed version of semantic-release that was changed in the package.json.

> If yes, how?

running `npm install` locally against your branch should take care of it. since the lockfile is currently using the version 1 format, it would be preferred if you could use a version of npm that is less than v7

greg0ire:
Done?

travi:
> Done?

getting really close. we still need to include the `ls-engines` check in the workflow, like https://github.com/semantic-release/github/pull/410/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R40-R41 and address my comment about adjusting the peer dependency range. with those two changes, i think we will be good to go.

travi (approved):
thanks a lot!

Issue #174, opened 2020-11-20
Use GitLab generic packages repository for assets


walkafwalka:
The [GitLab generic packages repository](https://docs.gitlab.com/ee/user/packages/generic_packages/index.html) is more suited for storing assets related to releases.

And unlike the current project uploads API, the [documentation explicitly states that the CI_JOB_TOKEN has access to the Packages API](https://docs.gitlab.com/ee/api/README.html#gitlab-ci-job-token).

nfriend:
@walkafwalka Thanks for this! I agree that the Generic Package Registry is a better fit. In fact, this is the direction GitLab itself is taking (see https://gitlab.com/gitlab-org/gitlab/-/issues/255271#note_453907441 and https://gitlab.com/gitlab-org/release-cli/-/issues/68).

I would suggest we wait until GitLab has landed on an specific approach and then updating this plugin to mirror its behavior.

mikaelkaron:
I think this has happened now: https://docs.gitlab.com/ee/user/project/releases/#use-a-generic-package-for-attaching-binaries

mfoti:
I have also an error uploading file with size greater than 100Mb:
`[semantic-release] [@semantic-release/gitlab] › ✖  An error occurred while uploading /builds/[...]/MyPackage-1.0.0-amd64.deb to the GitLab project uploads API:`
Can I try to post a PR?

fgreinacher:
This is now possible, thanks to @JonasSchubert work in #343 🎉

PR #283, opened 2021-10-05
feat: allow generic packages as assets #174

JonasSchubert:
Gitlab allows to add generic packages as assets attached to a release. #174 This is a suggestion to define a list of generics to be published into the generics repository of a project and then to be used as assets for the release.

Similar to `assets`, `generics` can be either an array of strings or an array of objects.
I would limit it to complete file paths (no globs), as (in my understanding) generic packages are limited to one file. (Please correct me if I'm wrong here)

Furthermore `generics` objects can have the properties, `path`, `label` (used for the file path in the generics package URL) and a `status` (default or hidden package).

I tried to avoid code duplications as good as possible by creating external functions. Could have avoided more, but stopped to maintain readability.

Additionally I added the possibility to debug unit tests in VS Code. I can remove this though, if no editor code should be added in this project.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,21 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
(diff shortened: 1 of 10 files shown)


max-petrovich:
Please reopen the PR, I think it would be very helpful!

PR #310, opened 2021-12-14
feat: add HTTP(S) proxy support

b3nk3:
This PR proposes the implementation of a http/https proxy support based on [this PR](https://github.com/semantic-release/gitlab/pull/229)

As pointed out by @alexjfischer on the above PR some extra changes had to be made around gitlab url props, which is included.

This is not my original work, rather a rebase and fixup of the original PR.

I would also like to point out, that we have been running this solution for over 6 months in our production environment.

@travi has raised valid points on the other PR about not taking on tech debt, and recommended the same approach as the GitHub plugin, however based on the extra dependencies and therefore peer dependencies, this implementation is cleaner and poses less risk in support and future tech debt. [More info](https://github.com/semantic-release/gitlab/pull/229#issuecomment-993592863)

Thanks in advance for reviewing. Do let me know if I can be of assistance to get this merged in a timely manner.

PR #316, opened 2022-01-06
feat: Swap got for fetch and adds proxy support

naffers:
An attempt at visualising what changes would be required to align the HTTP client with added proxy support. [As discussed in this thread](https://github.com/semantic-release/gitlab/pull/229#issuecomment-985820194).

There's a fair few differences between the `got` and `fetch` HTTP clients, which has resulted in what I'd consider some less than desirable changes.

Though hopefully this helps visualise some way the extent of the changes being suggested.

For reference the original PR: #229 
b3nk3 PR that refreshes the above: #310

### DIFF
--- a/lib/got-pollyfill-error.js
+++ b/lib/got-pollyfill-error.js
@@ -1,0 +1,8 @@
+module.exports = class GotPollyFillError extends Error {
+  constructor(message, statusCode) {
+    super(message);
+    this.name = 'GotPollyFillError';
+    this.response = {};
+    this.response.statusCode = statusCode;
+  }
+};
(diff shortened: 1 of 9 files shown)


test/verify.test.js
```diff
@@ -509,7 +509,7 @@ test.serial('Throw error if GitLab API return any other errors', async t => {
   const env = {GITLAB_TOKEN: 'gitlab_token'};
   const gitlab = authenticate(env)
     .get(`/projects/${owner}%2F${repo}`)
-    .times(3)
+    .times(1)
```
naffers, line 512: This was changed intentionally so as to not bloat the diffs with additional logic to facilitate what I can only assume was default retry logic contains in the `got` module.

lib/got-pollyfill-error.js
```diff
@@ -0,0 +1,8 @@
+module.exports = class GotPollyFillError extends Error {
```
naffers, line 1: This feels dirty but felt like the path of least resistance without rewriting the error handling

fgreinacher:
Closing this in favor of #310, thanks for your work so far @naffers!

PR #332, opened 2022-02-04
feat: comment on issues and merge requests resolved by current release

### LINKED ISSUE #40: Comment on issues resolved by current release
This may be on a roadmap but I didn't see any issues related to it so I figured I'd drop one here to at least indicate my support:

Similar to `semantic-release/github`, it would be awesome if the gitlab package had success/fail plugins to comment on issues related to the current release.

---

As an aside If I can get buy-in from my manager to work on this I'd be happy to but it's nothing to bank on

fgreinacher:
BREAKING CHANGE: By default all related issues and MRs are commented. Set the `successComment` option to `false` to disable that.

Fixes #40

🛠 with ❤ at [Siemens](https://opensource.siemens.io)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
 | `gitlabApiPathPrefix` | The GitLab API prefix.                                                                                                                         | `GL_PREFIX` or `GITLAB_PREFIX` environment variable or CI provided environment variables if running on [GitLab CI/CD](https://docs.gitlab.com/ee/ci) or `/api/v4`.      |
 | `assets`              | An array of files to upload to the release. See [assets](#assets).                                                                             | -                                                                                                                                                                       |
 | `milestones`          | An array of milestone titles to associate to the release. See [GitLab Release API](https://docs.gitlab.com/ee/api/releases/#create-a-release). | -                                                                                                                                                                       |
+| `postComments`          | Post comments to issues and MRs associated with a release. | `false`                                                                                                                                                                       |
 
 #### assets
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # @semantic-release/gitlab
+# @semantic-release/gitlab
 
 [**semantic-release**](https://github.com/semantic-release/semantic-release) plugin to publish a
 [GitLab release](https://docs.gitlab.com/ee/user/project/releases/).
@@ -10,6 +11,7 @@
 |--------------------|-----------------------------------------------------------------------------------------------------------------------|
 | `verifyConditions` | Verify the presence and the validity of the authentication (set via [environment variables](#environment-variables)). |
 | `publish`          | Publish a [GitLab release](https://docs.gitlab.com/ee/user/project/releases/).                                        |
+| `success`          | Add a comment to each GitLab Issue or Merge Request resolved by the release.                                          |
 
 ## Install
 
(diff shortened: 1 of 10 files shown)


fgreinacher:
OK, this is ready for review.

See https://gitlab.com/fgreinacher/semantic-release-comments/-/merge_requests/4 and it's linked issues for a sample run from the current version of the this PR.

/cc @pvdlg @tmeijn @nfriend @joshdcomp as you have been active in the original issue.

README.md
```diff
@@ -64,6 +64,7 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
...
 | `assets`              | An array of files to upload to the release. See [assets](#assets).                                                                             | -                                                                                                                                                                       |
 | `milestones`          | An array of milestone titles to associate to the release. See [GitLab Release API](https://docs.gitlab.com/ee/api/releases/#create-a-release). | -                                                                                                                                                                       |
+| `postComments`          | Post comments to issues and MRs associated with a release. | `false`                                                                                                                                                                       |
```
fgreinacher, line 67: I went for the fully backwards compatible way here, but we might also default to `true`?

tmeijn:
@fgreinacher Awesome!

Two points of feedback I have:
1. The issue that the Merge Request closes does necessarily lives in the same project that the Merge Request is merged in. Looking at code, it does not account for that case.
2. Related to 1: the release is also not necessarily in the same project, so constructing a full path should solve that case.

Thanks again!

fgreinacher:
> @fgreinacher Awesome!
> 
> Two points of feedback I have:
> 
> 1. The issue that the Merge Request closes does necessarily lives in the same project that the Merge Request is merged in. Looking at code, it does not account for that case.
> 2. Related to 1: the release is also not necessarily in the same project, so constructing a full url path should solve that case.
> 
> Thanks again!

Good catch, thanks! I fixed this and also use the full release URL now.

fgreinacher:
@semantic-release I am not sure who is currently maintaining this repository. Could you help me find the right reviewer?

fgreinacher:
@travi Are you currently active here and would review this? 🙏

README.md
travi: I hesitate slightly here because this differs from the [options available in the github plugin](https://github.com/semantic-release/github/#options). i would ideally like the functionality between the two plugins to be as similar as possible. 

that plugin currently allows changing the templated messages for `successComment` and `failComment`, allowing the comments to be disabled by setting the options to `false`. while i don't necessarily like that the templated approach would be a more complicated implementation, i do like the alignment between the two plugins with that approach.

travi: i like that defaulting to `false` with this implementation could make this available without it being a breaking change. if we switched from `postComments` to an approach that exposed the `successComment` and `failComment`, i think defaulting to `false` would be more difficult to justify. however, enabling the comments by default would be unexpected to existing users and likely not desired by at least some consumers of the plugin. 

i think i would lean toward enabling the comments by default if we do decide to go with that approach, but making it a breaking change.

test/resolve-config.test.js
```diff
@@ -35,6 +40,7 @@ test('Returns user config via environment variables', t => {
       gitlabApiUrl: urlJoin(gitlabUrl, gitlabApiPathPrefix),
       assets,
       milestones,
+      postComments: false,
```
travi, line 43: if the default value of this option is `false`, do all of these tests need to set explicitly to `false`? i think it could be good to have at least one test that proves it can handle being set to `false`, but it seems appropriate to leave most of the existing tests with the value unset.

test/resolve-config.test.js
```diff
@@ -17,7 +17,7 @@ test('Returns user config', t => {
       gitlabApiUrl: urlJoin(gitlabUrl, gitlabApiPathPrefix),
       assets,
       milestones: undefined,
-      postComments,
+      successComment: undefined,
```
travi, line 20: do these still need to be explicitly set to `undefined` here?

fgreinacher: Yes, because the property will be there, even if not passed via properties, and the tests do a deep equality comparison.

test/resolve-config.test.js
fgreinacher: See https://github.com/semantic-release/gitlab/pull/332#discussion_r830478105 :)

README.md
fgreinacher: All done, happy to get some feedback @travi!

travi (approved):

fgreinacher:
@travi Eek, seems the squash commit did not contain my breaking change trailer, making this 7.1.0 instead of 8.0.0 :/ Should I revert this for 7.1.1 and recommit with the correct message?

travi:
Yeah, that's the route that I would recommend. I've approved your draft PR for the revert. 

I do wish that GitHub would do better with copying commit body details into the suggested squash message in situations like this.

PR #322, opened 2022-01-23
feat: allow specify url in release assets

awcjack:
This solve https://github.com/semantic-release/gitlab/issues/154

PR #343, opened 2022-03-22
feat(generics): allow upload of build assets as generic packages #174

JonasSchubert:
> Progress of #283 which I closed due to missing time on my side

Gitlab allows to add generic packages as assets attached to a release. #174 This is a suggestion to define a list of generics to be published into the generics repository of a project and then to be used as assets for the release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The plugin can be configured in the [**semantic-release** configuration file](ht
         "gitlabUrl": "https://custom.gitlab.com",
         "assets": [
           { "path": "dist/asset.min.css", "label": "CSS distribution" },
-          { "path": "dist/asset.min.js", "label": "JS distribution" },
+          { "path": "dist/asset.min.js", "label": "JS distribution", "type": "generic_package" },
           { "path": "dist/asset.min.js", "label": "v${nextRelease.version}.js" },
           { "url": "https://gitlab.com/gitlab-org/gitlab/-/blob/master/README.md" }
         ]
@@ -106,6 +106,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` 
 | `label`    | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file. Supports [Lodash templating](https://lodash.com/docs#template).                                                                                                                                                       | File name extracted from the `path`. |
 | `type`     | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). Supports [Lodash templating](https://lodash.com/docs#template).                                                       | `other`                              |
 | `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. Supports [Lodash templating](https://lodash.com/docs#template). | -                                    |
+| `target`   | Controls where the file is uploaded to. Can be set to `project_upload` for storing the file as [project upload](https://docs.gitlab.com/ee/api/projects.html#upload-a-file) or `generic_package` for storing the file as [generic package](https://docs.gitlab.com/ee/user/packages/generic_packages/).                            | `project_upload`                     |
+| `status`   | This is only applied, if `target` is set to `generic_package`. The generic package status. Can be `default` and `hidden` (see official documents on [generic packages](https://docs.gitlab.com/ee/user/packages/generic_packages/)).                                                                                               | `default`                            |
 
 Each entry in the `assets` `Array` is globbed individually. A [glob](https://github.com/isaacs/node-glob#glob-primer)
 can be a `String` (`"dist/**/*.js"` or `"dist/mylib.js"`) or an `Array` of `String`s that will be globbed together
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -50,6 +50,9 @@ export default async (pluginConfig, context) => {
         const label = asset.label ? template(asset.label)(context) : undefined;
         const type = asset.type ? template(asset.type)(context) : undefined;
         const filepath = asset.filepath ? template(asset.filepath)(context) : undefined;
+        const target = asset.target ? template(asset.target)(context) : undefined;
+        const status = asset.status ? template(asset.status)(context) : undefined;
+
         if (_url) {
           assetsList.push({ label, rawUrl: _url, type, filepath });
           debug("use link from release setting: %s", _url);
@@ -74,28 +77,59 @@ export default async (pluginConfig, context) => {
           debug("file label: %o", label);
           debug("file type: %o", type);
           debug("file filepath: %o", filepath);
+          debug("file target: %o", target);
+          debug("file status: %o", status);
 
           // Uploaded assets to the project
           const form = new FormData();
           form.append("file", createReadStream(file));
 
-          const uploadEndpoint = urlJoin(gitlabApiUrl, `/projects/${encodedRepoId}/uploads`);
-
-          debug("POST-ing the file %s to %s", file, uploadEndpoint);
-
+          let uploadEndpoint;
           let response;
-          try {
-            response = await got.post(uploadEndpoint, { ...apiOptions, ...proxy, body: form }).json();
-          } catch (error) {
-            logger.error("An error occurred while uploading %s to the GitLab project uploads API:\n%O", file, error);
-            throw error;
+
+          if (target === "generic_package") {
+            // Upload generic packages
+            const encodedLabel = encodeURIComponent(label);
+            uploadEndpoint = urlJoin(
+              gitlabApiUrl,
+              `/projects/${encodedRepoId}/packages/generic/release/${encodedGitTag}/${encodedLabel}?${
+                status ? `status=${status}&` : ""
+              }select=package_file`
+            );
+
+            debug("PUT-ing the file %s to %s", file, uploadEndpoint);
+
+            try {
+              response = await got.put(uploadEndpoint, { ...apiOptions, body: form }).json();
+            } catch (error) {
+              logger.error("An error occurred while uploading %s to the GitLab generics package API:\n%O", file, error);
+              throw error;
+            }
+
+            const { url } = response.file;
+
+            assetsList.push({ label, alt: "release", url, type: "package", filepath });
+
+            logger.log("Uploaded file: %s", url);
+          } else {
+            // Handle normal assets
+            uploadEndpoint = urlJoin(gitlabApiUrl, `/projects/${encodedRepoId}/uploads`);
+
+            debug("POST-ing the file %s to %s", file, uploadEndpoint);
+
+            try {
+              response = await got.post(uploadEndpoint, { ...apiOptions, ...proxy, body: form }).json();
+            } catch (error) {
+              logger.error("An error occurred while uploading %s to the GitLab project uploads API:\n%O", file, error);
+              throw error;
+            }
+
+            const { url, alt } = response;
+
+            assetsList.push({ label, alt, url, type, filepath });
+
+            logger.log("Uploaded file: %s", url);
           }
-
-          const { url, alt } = response;
-
-          assetsList.push({ label, alt, url, type, filepath });
-
-          logger.log("Uploaded file: %s", url);
         }
       })
     );
--- a/test/publish.test.js
+++ b/test/publish.test.js
@@ -83,6 +83,50 @@ test.serial("Publish a release with assets", async (t) => {
   t.true(gitlab.isDone());
 });
 
+test.serial("Publish a release with generics", async (t) => {
+  const cwd = "test/fixtures/files";
+  const owner = "test_user";
+  const repo = "test_repo";
+  const env = { GITLAB_TOKEN: "gitlab_token" };
+  const nextRelease = { gitHead: "123", gitTag: "v1.0.0", notes: "Test release note body" };
+  const options = { repositoryUrl: `https://gitlab.com/${owner}/${repo}.git` };
+  const encodedRepoId = encodeURIComponent(`${owner}/${repo}`);
+  const encodedGitTag = encodeURIComponent(nextRelease.gitTag);
+  const uploaded = { file: { url: "/uploads/file.css" } };
+  const generic = { path: "file.css", label: "Style package", target: "generic_package", status: "hidden" };
+  const assets = [generic];
+  const encodedLabel = encodeURIComponent(generic.label);
+  const gitlab = authenticate(env)
+    .post(`/projects/${encodedRepoId}/releases`, {
+      tag_name: nextRelease.gitTag,
+      description: nextRelease.notes,
+      assets: {
+        links: [
+          {
+            name: "Style package",
+            url: `https://gitlab.com/${owner}/${repo}${uploaded.file.url}`,
+            link_type: "package",
+          },
+        ],
+      },
+    })
+    .reply(200);
+  const gitlabUpload = authenticate(env)
+    .put(
+      `/projects/${encodedRepoId}/packages/generic/release/${encodedGitTag}/${encodedLabel}?status=${generic.status}&select=package_file`,
+      /filename="file.css"/gm
+    )
+    .reply(200, uploaded);
+
+  const result = await publish({ assets }, { env, cwd, options, nextRelease, logger: t.context.logger });
+
+  t.is(result.url, `https://gitlab.com/${owner}/${repo}/-/releases/${encodedGitTag}`);
+  t.deepEqual(t.context.log.args[0], ["Uploaded file: %s", uploaded.file.url]);
+  t.deepEqual(t.context.log.args[1], ["Published GitLab release: %s", nextRelease.gitTag]);
+  t.true(gitlabUpload.isDone());
+  t.true(gitlab.isDone());
+});
+
 test.serial("Publish a release with asset type and permalink", async (t) => {
   const cwd = "test/fixtures/files";
   const owner = "test_user";


README.md
```diff
@@ -102,6 +106,24 @@ distribution` and `MyLibrary CSS distribution` in the GitLab release.
 `css` files in the `dist` directory and its sub-directories excluding the minified version, plus the
 `build/MyLibrary.zip` file and label it `MyLibrary` in the GitLab release.
 
+#### generics
```
fgreinacher, line 109: I'm wondering whether it might be simpler for users if we would add this functionality to `assets` instead?

```json
assets: [
   { path: 'foo.txt', label: 'Foo', type: 'generic_package' },
   { path: 'bar.txt', label: 'Bar', type: 'project_upload' }
]
```

JonasSchubert: Sounds like a better solution.
Should we define the default value for `type` to `project_upload` to avoid breaking changes or should this be implemented as a breaking change.
Here I would vote against breaking changes here. 😄

fgreinacher: Yes, I agree that we should not make this a breaking change, thanks for taking care!

fgreinacher:
@JonasSchubert Could you check if it makes sense to pull the changes from #322 in this MR? It seems to touch the very same code paths.

JonasSchubert:
> @JonasSchubert Could you check if it makes sense to pull the changes from #322 in this MR? It seems to touch the very same code paths.

I would tackle this in a separate PR. I checked the code and in my opinion it needs major adjustments regarding assets handling.

README.md
```diff
@@ -95,8 +95,9 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
 | -------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
 | `path`   | **Required.** A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
 | `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file.| File name extracted from the `path`. |
-| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). | `other` |
+| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). If this is set to `generic_package` it will upload the `glob` as a [generic package](https://docs.gitlab.com/ee/user/packages/generic_packages/). | `other` |
```
fgreinacher, line 98: I like the simplicity of reusing the existing option! I don't like overloading the semantics of this option. I think I'd prefer a separate option (maybe `storage` with possible values `project_upload` and `generic_package`). What's your opinion @JonasSchubert?

lib/publish.js
```diff
@@ -56,28 +56,58 @@ module.exports = async (pluginConfig, context) => {
...
-        } catch (error) {
-          logger.error('An error occurred while uploading %s to the GitLab project uploads API:\n%O', file, error);
-          throw error;
-        }
+          debug('PUT-ing the file %s to %s', file, uploadEndpoint);
+
+          try {
+            response = await got.put(uploadEndpoint, {...apiOptions, body: form}).json();
+          } catch (error) {
+            logger.error('An error occurred while uploading %s to the GitLab generics package API:\n%O', file, error);
```
fgreinacher, line 83: Nit:

```suggestion
            logger.error('An error occurred while uploading %s to the GitLab generic packages API:\n%O', file, error);
```

README.md
JonasSchubert: Adding a new option would keep things in a clean way, indeed.
`type` values would only contain the documented [link types](https://docs.gitlab.com/ee/user/project/releases/#link-types).

JonasSchubert: Instead of `storage` maybe `target`. Also considering the URL feature?

fgreinacher: @JonasSchubert Yes, I like `target`!

README.md
```diff
@@ -96,8 +96,10 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
...
+| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)).  | `other` |
 | `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. | - |
+| `target` | If this is set to `project_upload` it will uppload the `glob` to the release. But if set to `generic_package` it will upload the `glob` as a [generic package](https://docs.gitlab.com/ee/user/packages/generic_packages/). Can be `project_upload` or `generic_package`. | `project_upload` |
+| `status` | This is only applied, if `type` is set to `generic_package`. The generic package status. Can be `default` and `hidden` (see official documents on [generic packages](https://docs.gitlab.com/ee/user/packages/generic_packages/)). | `default` |
```
fgreinacher, line 102:
```suggestion
| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)).  | `other` |
| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. | - |
| `target` | Controls where the file is uploaded to. Can be set to `project_upload` for storing the file as [project upload](https://docs.gitlab.com/ee/api/projects.html#upload-a-file) or `generic_package` for storing the file as [generic package](https://docs.gitlab.com/ee/user/packages/generic_packages/). | `project_upload` |
| `status` | This is only applied, if `target` is set to `generic_package`. The generic package status. Can be `default` and `hidden` (see official documents on [generic packages](https://docs.gitlab.com/ee/user/packages/generic_packages/)). | `default` |
```

fgreinacher (changes requested):
Again just a small documentation suggestion.

JonasSchubert:
I would wait for #322 and merge afterwards.

fgreinacher:
> I would wait for #322 and merge afterwards.

Fine for me! Ping me again in case you change your mind :)

mfoti:
any update on this?

bgoareguer:
@fgreinacher @JonasSchubert #322 has been merged. Can you please merge this PR?

fgreinacher:
Happy to merge once conflicts are resolved. @JonasSchubert are you still willing to work on this?

JonasSchubert:
Sorry everyone! I lost track of some stuff...
Fixed the conflicts and ready to merge. Will keep an eye in case any issues should occur

JonasSchubert:
@fgreinacher Rebased to v10 and fixed lint errors. Should be fine now.

fgreinacher (approved):
LGTM, let's get this in finally :)

mfoti:
Thanks for the work to all, and for merging this. 

I've tried this feature and I guess there are two bugs: 

1. the encoded url should end with the file name, not the label here: https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L91 as in documentation (https://docs.gitlab.com/ee/user/packages/generic_packages/#publish-a-package-file) 

The url should be like this:

`PUT /projects/:id/packages/generic/:package_name/:package_version/:file_name?status=:status`

I got this error creating labels like this: "MyApp x86 (.dmg)"

2. The url should not be taken from the response body (https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L105) but should be the value of `uploadEndpoint`. 
Actually when it does a release, it publish the package with an url like this: 

```
"links": [
      {
        "name": "MyApp-v2.9-x64.dmg",
        "url": "https://storage.googleapis.com/gitlab-gprd-package-repo/25/87/2587b9f64886c595ba69e1e67434ca7c69a120154ad2d3d5a4a10ddca78a978e/packages/12549696/files/68634567/MyApp-v2.9-x64.dmg?GoogleAccessId=gitlab-object-storage-prd@gitlab-production.iam.gserviceaccount.com&Signature=M03DQehdSKyLvpdTBKAuxiveI4yANOMgKcFs5etCp%2FTwGoj%2BoQy%2FRpoiSPGd%0Ak6leI4uVtOv55w%2BZa764o%2FENpOQ2eFTOfIr6H45Hyi5QupPArB0OL0164Wul%0AgmhqQtcP%2FZ%2FQEVm%2F40Vq%2Fa50MPoPgOa4%2BRt%2FDS9CjWVPXz2PvKz9Zzy0P114%0AXiEyaMxSCQmPcai6%2BYFetwTI%2Fi%2Fm4HaSkOqG%2BLhHd%2BswJDNZNBucZo7gf2fg%0APbGQkJ3wQxzVY%2BukhxjMm%2BjXw8kR%2F%2FPsu8F810gjLzMsnCYNt2O5pwpCY5fr%0AzDP%2BQ1YosqKiTuBhWR6kPUUh43j6mdKcjmbGv7vJcA%3D%3D&Expires=1676411480",
        "link_type": "package"
      }
```
After few minutes the package is no longer available, loading the package url I got this response:

```
<Error>
<Code>ExpiredToken</Code>
<Message>The provided token has expired.</Message>
<Details>Request signature expired at: 2023-02-14T21:51:20+00:00</Details>
</Error>
````

I guess that this can fix the bug: `assetsList.push({ label, alt: "release", uploadEndpoint, type: "package", filepath, target });`

I can do the PRs if needed

**Comment by JonasSchubert** [comment]
Hi @mfoti and thanks for your report. 

> 1. the encoded url should end with the file name, not the label here: https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L91 as in documentation (https://docs.gitlab.com/ee/user/packages/generic_packages/#publish-a-package-file) 
> 
> The url should be like this:
> 
> `PUT /projects/:id/packages/generic/:package_name/:package_version/:file_name?status=:status`
> 
> This code should fix: `/projects/${encodedRepoId}/packages/generic/release/${encodedGitTag}/${encodedFileName}`
> 
> I got this error creating labels like this: "MyApp x86 (.dmg)"

As the [README](https://github.com/semantic-release/gitlab#assets) des describes, the `label` represents the file name. IMO the current implementation is here correct. 

> 2. The url should not be taken from the response body (https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L105) but should be the value of `uploadEndpoint`. 
> Actually when it does a release, it publish the package with an url like this: 
> 
> ```
> "links": [
>       {
>         "name": "MyApp-v2.9-x64.dmg",
>         "url": "https://storage.googleapis.com/gitlab-gprd-package-repo/25/87/2587b9f64886c595ba69e1e67434ca7c69a120154ad2d3d5a4a10ddca78a978e/packages/12549696/files/68634567/MyApp-v2.9-x64.dmg?GoogleAccessId=gitlab-object-storage-prd@gitlab-production.iam.gserviceaccount.com&Signature=M03DQehdSKyLvpdTBKAuxiveI4yANOMgKcFs5etCp%2FTwGoj%2BoQy%2FRpoiSPGd%0Ak6leI4uVtOv55w%2BZa764o%2FENpOQ2eFTOfIr6H45Hyi5QupPArB0OL0164Wul%0AgmhqQtcP%2FZ%2FQEVm%2F40Vq%2Fa50MPoPgOa4%2BRt%2FDS9CjWVPXz2PvKz9Zzy0P114%0AXiEyaMxSCQmPcai6%2BYFetwTI%2Fi%2Fm4HaSkOqG%2BLhHd%2BswJDNZNBucZo7gf2fg%0APbGQkJ3wQxzVY%2BukhxjMm%2BjXw8kR%2F%2FPsu8F810gjLzMsnCYNt2O5pwpCY5fr%0AzDP%2BQ1YosqKiTuBhWR6kPUUh43j6mdKcjmbGv7vJcA%3D%3D&Expires=1676411480",
>         "link_type": "package"
>       }
> ```
> After few minutes the package is no longer available, loading the package url I got this response:
> 
> ```
> <Error>
> <Code>ExpiredToken</Code>
> <Message>The provided token has expired.</Message>
> <Details>Request signature expired at: 2023-02-14T21:51:20+00:00</Details>
> </Error>
> ````
> 
> I guess that this can fix the bug: `assetsList.push({ label, alt: "release", uploadEndpoint, type: "package", filepath, target });`

This is a little bit more tricky then `1.`.

For me it looks like you use Google Storage to store some assets and packages and this provider returns an URL with an expiration date. Not sure whether this is for the URL only or also for the file.
Interesting would be what happens if we use the `uploadEndpoint`? Even if we use this URL isn't the returned URL still active in the background and by this probably causing the same issue? 
Would it be possible for you to test this in your project? 

BTW the currently used URL is the one provided by the API call response according to the docs. 
With this said, the docs also highlight [how to download a generic package] (https://docs.gitlab.com/ee/user/packages/generic_packages/#download-package-file). 

But:
> If multiple packages have the same name, version, and filename, then the most recent one is retrieved.

With the current approach we have direct access to the uploaded file without worrying about this drawback, which we might loose using the Gitlab URL. 

@mfoti what are your thoughts on this?

**Comment by mfoti** [comment]
> As the [README](https://github.com/semantic-release/gitlab#assets) describes, the `label` represents the file name. IMO the current implementation is here correct.

I guess defaults to the filename, but allow to put a string as package description, as in my case: "MacOS x64 (.dmg)" 

BTW I can confirm that using the filename as label works: 

```
       // This works
       {
          label: `MyApp-${nextRelease.version}-x64.dmg`,
          path: `release/build/MyApp-${nextRelease.version}-x64.dmg`,
          type: "image",
          target: "generic_package"
        },
```

But if you use a different label it doesn't

```
       // This doesn't works
       {
          label: 'MacOS x64 (.dmg)',
          path: `release/build/MyApp-${nextRelease.version}-x64.dmg`,
          type: "image",
          target: "generic_package"
        },
```

I guess to avoid errors (release is correctly generated with a label different of the filename using the default target) we can just ignore the label in the code and use the filename here:

https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L88 

IMO is better to give the ability to users to use their own label. BTW This is not a blocking behavior, when the target is to `generic_package` using the filename as label create the release correctly.

> This is a little bit more tricky then `1.`.

yep man, it is lol 🤣 

> 
> For me it looks like you use Google Storage to store some assets and packages and this provider returns an URL with an expiration date. Not sure whether this is for the URL only or also for the file. Interesting would be what happens if we use the `uploadEndpoint`? Even if we use this URL isn't the returned URL still active in the background and by this probably causing the same issue? Would it be possible for you to test this in your project?

yes, I already tested it and it works :) I have a fork of the project with this fix and I can confirm resolve the issue. 

> 
> BTW the currently used URL is the one provided by the API call response according to the docs. With this said, the docs also highlight [how to download a generic package](https://docs.gitlab.com/ee/user/packages/generic_packages/#download-package-file).

Yes you are right, but I guess is not well documented, I'll try to find the code on gitlab repo to check better the behavior. What can I say is that my builds are greater the 200mb, probably gitlab uses google storage if the size of the package is bigger than a certain value. I don't know. But the url provided in filename is broken in my case, and this is a blocking bug that doesn't allow me to use this kind of release 

> 
> But:
> 
> > If multiple packages have the same name, version, and filename, then the most recent one is retrieved.
> 
> With the current approach we have direct access to the uploaded file without worrying about this drawback, which we might loose using the Gitlab URL.

To be honest I don't understand which is this case, let's assume I'll build two different packages with the same name, for example: 

- /build/x86/MyApp.dmg <-for x86
- /build/arm64/MyApp.dmg <-for arm64

And try to publish them. In this case I will have an url like: 
/projects/${encodedRepoId}/packages/generic/release/my-tag-v1.0/MyApp.dmg 

That of course will store the last published file to that path. At the end you will always get the most recent (last published) package downloading the content from the package path

The main problem is that it doesn't works for me as is now, I have to use my fork.

**Comment by mfoti** [comment]
I guess there is another feature that can be used too, the package type.

At the moment it is hardcoded: 

`assetsList.push({ label, alt: "release", url, type: "package", filepath, target });` (https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L107) 

This type value should be taken from the configuration, this should fix: 
```
const type = asset.type ? template(asset.type)(context) : "package";
assetsList.push({ label, alt: "release", url, type, filepath, target });
```

The final effect, without hardcoding the type, will be this:
Screen Shot 2023-02-15 at 12 04 03

JonasSchubert:
@mfoti You already created and fork with a fix on your side. Feel free to create the PR ;)
Already thanks for your work!

mfoti:
I should squash, but is a really easy fix: https://github.com/semantic-release/gitlab/pull/504, this fixes only the main "bug", but not the package type and the filename/label

JonasSchubert:
> I guess there is another feature that can be used too, the package type.
> 
> At the moment it is hardcoded:
> 
> `assetsList.push({ label, alt: "release", url, type: "package", filepath, target });` (https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L107)

IMO the current implementation is correct here. Generic packages only should have the type `packages`. The other `type` options (`runbook`, `image` and `other`) are for assets and not packages.
This can be used with either  not setting the property `target` at all or setting the (default) value `project_upload`.

JonasSchubert:
> I guess defaults to the filename, but allow to put a string as package description, as in my case: "MacOS x64 (.dmg)"
> 
> BTW I can confirm that using the filename as label works:
> 
> ```
>        // This works
>        {
>           label: `MyApp-${nextRelease.version}-x64.dmg`,
>           path: `release/build/MyApp-${nextRelease.version}-x64.dmg`,
>           type: "image",
>           target: "generic_package"
>         },
> ```
> 
> But if you use a different label it doesn't
> 
> ```
>        // This doesn't works
>        {
>           label: 'MacOS x64 (.dmg)',
>           path: `release/build/MyApp-${nextRelease.version}-x64.dmg`,
>           type: "image",
>           target: "generic_package"
>         },
> ```
> 
> I guess to avoid errors (release is correctly generated with a label different of the filename using the default target) we can just ignore the label in the code and use the filename here:
> 
> https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L88
> 
> IMO is better to give the ability to users to use their own label. BTW This is not a blocking behavior, when the target is to `generic_package` using the filename as label create the release correctly.

I think this is a create idea. Would you like to add it to your PR too?

PR #349, opened 2022-03-23
feat: create issues when the release is failing

JonasSchubert:
- adapted from [github](https://github.com/semantic-release/github)

### DIFF
--- a/lib/fail.js
+++ b/lib/fail.js
@@ -1,0 +1,60 @@
+const {template} = require('lodash');
+const urlJoin = require('url-join');
+const got = require('got');
+const debug = require('debug')('semantic-release:gitlab');
+const resolveConfig = require('./resolve-config');
+const getRepoId = require('./get-repo-id');
+const getFailComment = require('./get-fail-comment');
+
+module.exports = async (pluginConfig, context) => {
+  const {
+    options: {repositoryUrl},
+    branch,
+    errors,
+    logger,
+  } = context;
+  const {gitlabToken, gitlabUrl, gitlabApiUrl, failComment, failTitle, labels, assignee} = resolveConfig(
+    pluginConfig,
+    context
+  );
+  const repoId = getRepoId(context, gitlabUrl, repositoryUrl);
+  const encodedRepoId = encodeURIComponent(repoId);
+  const apiOptions = {headers: {'PRIVATE-TOKEN': gitlabToken}};
+
+  if (failComment === false || failTitle === false) {
+    logger.log('Skip issue creation.');
+  } else {
+    const description = failComment ? template(failComment)({branch, errors}) : getFailComment(branch, errors);
+
+    const issuesEndpoint = urlJoin(gitlabApiUrl, `/projects/${repoId}/issues`);
+    const openIssueEndpoint = urlJoin(issuesEndpoint, '?state=opened');
+    const openIssues = await got(openIssueEndpoint, {...apiOptions}).json();
+    const existingIssue = openIssues.find(openIssue => openIssue.title === failTitle);
+
+    if (existingIssue) {
+      debug('comment on issue: %O', existingIssue);
+
+      const issueNotesEndpoint = urlJoin(
+        gitlabApiUrl,
+        `/projects/${existingIssue.project_id}/issues/${existingIssue.iid}/notes`
+      );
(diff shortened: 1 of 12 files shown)


test/resolve-config.test.js
```diff
@@ -8,16 +8,21 @@ test('Returns user config', t => {
...
       gitlabUrl,
       gitlabApiUrl: urlJoin(gitlabUrl, gitlabApiPathPrefix),
       assets,
       milestones: undefined,
       successComment: undefined,
+      failTitle: 'The automated release is failing 🚨',
+      failComment: undefined,
+      labels: false,
+      assignee: undefined,
     }
```
fgreinacher, line 26: Maybe we should define an object with the default values and reuse that across all test cases to reduce duplication?

Something like this:

```js
const defaultOptions = {
      // ...
      successComment: undefined,
      failTitle: 'The automated release is failing 🚨',
      assignee: undefined,
      // ...
}

// ...

t.deepEqual(
    resolveConfig({gitlabUrl, gitlabApiPathPrefix, assets, postComments, labels}, {env: {GITLAB_TOKEN: gitlabToken}}),
    {
      ... defaultOptions,
      gitlabApiUrl: urlJoin(gitlabUrl, gitlabApiPathPrefix),
    }
```

fgreinacher (changes requested):
Thanks @JonasSchubert, this is looking great! Just a small idea to reduce duplication in tests.

lib/fail.js
```diff
@@ -0,0 +1,60 @@
...
+  const encodedRepoId = encodeURIComponent(repoId);
+  const apiOptions = {headers: {'PRIVATE-TOKEN': gitlabToken}};
+
+  if (failComment === false || failTitle === false) {
+    logger.log('Skip issue creation.');
+  } else {
+    const description = failComment ? template(failComment)({branch, errors}) : getFailComment(branch, errors);
+
+    const issuesEndpoint = urlJoin(gitlabApiUrl, `/projects/${repoId}/issues`);
+    const openIssueEndpoint = urlJoin(issuesEndpoint, '?state=opened&per_page=100');
```
JonasSchubert, line 30: We have projects with a huge amount of issues. The Gitlab API returns by default only the [first 20 issues](https://docs.gitlab.com/ee/api/#offset-based-pagination). This is why I increased the `per_page` size to the `max` of `100`.
Is this OK or should we use the default size. Or should we even search through all pages?

(Might also be relevant for the `success`-hook)

fgreinacher: Oh, that's a good finding! 

I don't like the `100` either because it just defers the problem.

What about using the `search` parameter to prefilter for the expected title: https://docs.gitlab.com/ee/api/issues.html?msclkid=31d7fcfdb4b211ec99728d2bf9d701dc#list-project-issues

fgreinacher (approved):

jaklan:
@JonasSchubert it would be worth to add some example how to pass multiple labels, especially when they contain whitespaces, because it's pretty non-obvious without digging into code.

JonasSchubert:
Would it be helpful to link the API definition?
https://docs.gitlab.com/ee/api/issues.html#new-issue

jaklan:
@JonasSchubert tbh not really, I had to to my own test, because that case was not covered by docs.

What I discovered is:
a) `+` instead of spaces work:
```
curl --header "PRIVATE-TOKEN: $GITLAB_TOKEN" "https://code.roche.com/api/v4/issues?labels=0)+Release"
```
b) `%20` instead of spaces work:
```
curl --header "PRIVATE-TOKEN: $GITLAB_TOKEN" "https://code.roche.com/api/v4/issues?labels=0)%20Release"
```
Labels should be separated just with comma, if you want to include whitespace after comma - it also has to be encoded as `+` or `%20` - so both works:
```
curl --header "PRIVATE-TOKEN: $GITLAB_TOKEN" "https://code.roche.com/api/v4/issues?labels=0)+Release,3)+Type:+Bug"
curl --header "PRIVATE-TOKEN: $GITLAB_TOKEN" "https://code.roche.com/api/v4/issues?labels=0)+Release,+3)+Type:+Bug"
```
I haven't tested that with `semantic-release` yet, but I assume it should work. The question is what would happen if I leave spaces instead of `+`/`%20` - would it be encoded automatically or not?

Issue #344, opened 2022-03-22
successComment has bad formatted emojis


baby-gnu:
Hello.

There should be a space between `:package:` and `:rocket` because they are not displayed on my 14.6 self hosted gitlab.

![semantic-release-gitlab-successComment-emoji-fail](https://user-images.githubusercontent.com/1233212/159498495-b0a7a52c-1a09-45ed-9f7e-644cff71c846.png)

Thanks.

TilmannBach:
there is another bug in the comment: the link to the gitlab release does not work correctly:
![image](https://user-images.githubusercontent.com/2125941/159913503-0cd7ecaf-7d04-49cf-b7a4-cd6afb5e665c.png)
this `%2F` should be a slash instead

JonasSchubert:
#342 would fix both issues

PR #342, opened 2022-03-22
fix: display default icons and fix URL in success message

JonasSchubert:
First of all thanks to @fgreinacher for this nice feature.
We are using it already and I stumpled upon two issues:
- the :package: and :rocket: icon were not proberly displayed as they were not separated
- the URL to the release did not work because of `%2F` in the URI; this is why I added the `decodeURIComponent`

Fixes #344

PR #350, opened 2022-03-24
fix: render emojis correctly in success comments

languitar:
Gitlab requires a space between emojis to render them correctly.

Fixes: #344


JonasSchubert:
Duplicate of #342

PR #359, opened 2022-04-07
feat: add support for NO_PROXY

Babbafett:
# Please check if the MR fulfills these requirements
- [x] the commit message follows conventional commits
- [x] all required tests for the changes have been added
- [x] docs have been added in readme
# What kind of change does this MR introduce?
- Add NO_PROXY environment variable support
# What is the current behavior?
- NO_PROXY environment variables is completely ignored
# What is the new behavior (if this is a feature change)?
- NO_PROXY environment variables can be used to bypass the proxy configuration via HTTPS_PROXY and/or HTTP_PROXY envrionment variables
# Does this MR introduce a breaking change?
No
# Other information
In some cases it requires a web proxy to reach resources through the internet and outside of a company network. This will work when the gitlab Instance is also accessible through the internet, but in some cases the gitlab instance is a self hosted one and is only be reachable within the same network.
In case of that you need a proxy configuration (e.g. through HTTP_PROXY and/or HTTPS_PROXY for your build process with @semantic-release/exec) it would fail to reach the gitlab instance as a combination of proxy and no_proxy is not possible at the moment.
This PR will introduce this option to allow a combination of having HTTP_PROXY/HTTPS_PROXY and NO_PROXY environment variables

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
 | `GL_URL` or `GITLAB_URL`       | The GitLab endpoint.                                      |
 | `GL_PREFIX` or `GITLAB_PREFIX` | The GitLab API prefix.                                    |
 | `HTTP_PROXY` or `HTTPS_PROXY`  | HTTP or HTTPS proxy to use.                               |
-| `NO_PROXY`                     | NO_PROXY bypass proxy hosts                               |
+| `NO_PROXY`                     | Patterns for which the proxy should be ignored. See [details below](#proxy-configuration). |
 
 #### Proxy configuration
 
@@ -69,11 +69,7 @@ If your proxy server requires authentication embed the username and password in 
 
 If your GitLab instance is exposed via plain HTTP (not recommended!) use `HTTP_PROXY` instead.
 
-#### Proxy bypass configuration
-
-You can configure a bypass option by setting `NO_PROXY` environment variable: `NO_PROXY=*.host.com, host.com`
-
-Depending on your gitlabUrl and a matching `NO_PROXY`it is bypassing your proxy configuration of the `HTTP_PROXY` or `HTTPS_PROXY` environment variable.
+If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environment variable: NO_PROXY=*.host.com, host.com`
 
 ### Options
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ If your proxy server requires authentication embed the username and password in 
 
 If your GitLab instance is exposed via plain HTTP (not recommended!) use `HTTP_PROXY` instead.
 
-If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environment variable: NO_PROXY=*.host.com, host.com`
+If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environment variable: `NO_PROXY=*.host.com, host.com`
 
 ### Options
 
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
 | `GL_URL` or `GITLAB_URL`       | The GitLab endpoint.                                      |
 | `GL_PREFIX` or `GITLAB_PREFIX` | The GitLab API prefix.                                    |
 | `HTTP_PROXY` or `HTTPS_PROXY`  | HTTP or HTTPS proxy to use.                               |
+| `NO_PROXY`                     | NO_PROXY bypass proxy hosts                               |
 
 #### Proxy configuration
 
@@ -67,6 +68,12 @@ You can configure a proxy server via the `HTTPS_PROXY` environment variable: `HT
 If your proxy server requires authentication embed the username and password in the URL: `HTTPS_PROXY=http://user:pwd@proxyurl.com:8080`
 
 If your GitLab instance is exposed via plain HTTP (not recommended!) use `HTTP_PROXY` instead.
+
+#### Proxy bypass configuration
+
+You can configure a bypass option by setting `NO_PROXY` environment variable: `NO_PROXY=*.host.com, host.com`
+
+Depending on your gitlabUrl and a matching `NO_PROXY`it is bypassing your proxy configuration of the `HTTP_PROXY` or `HTTPS_PROXY` environment variable.
 
 ### Options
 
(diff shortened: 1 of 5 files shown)


lib/resolve-config.js
```diff
@@ -44,18 +45,31 @@ module.exports = (
...
+function bypassProxy(NO_PROXY, gitlabUrl) {
+  if (NO_PROXY === '*') {
+    return true;
+  }
+  const hostName = gitlabUrl.split('://')[1].split(':')[0];
+  const noProxyEntries = NO_PROXY.split(',');
+  return noProxyEntries.some((proxyConfig) => {
+    return hostName.endsWith(proxyConfig.toLowerCase().trim().replace(/^\./, ""));
+  });
+}
```
fgreinacher, line 61: I think it's a little bit more complex than this. Maybe have a look at the implementation in the pretty common https://www.npmjs.com/package/proxy-from-env package.

lib/resolve-config.js
```diff
@@ -44,18 +45,31 @@ module.exports = (
...
   const sharedParameters = {
     keepAlive: true,
     keepAliveMsecs: 1000,
     maxSockets: 256,
     maxFreeSockets: 256,
     scheduling: 'lifo',
   };
+  if(bypassProxy(NO_PROXY, gitlabUrl)) {
+    return {};
+  }
```
fgreinacher, line 72: Nit: Maybe most this to the beginning of the function.

README.md
```diff
@@ -57,6 +57,7 @@ Create a [personal access token](https://docs.gitlab.com/ce/user/profile/persona
 | `GL_URL` or `GITLAB_URL`       | The GitLab endpoint.                                      |
 | `GL_PREFIX` or `GITLAB_PREFIX` | The GitLab API prefix.                                    |
 | `HTTP_PROXY` or `HTTPS_PROXY`  | HTTP or HTTPS proxy to use.                               |
+| `NO_PROXY`                     | NO_PROXY bypass proxy hosts                               |
```
fgreinacher, line 60:
```suggestion
| `NO_PROXY`                     | Patterns for which the proxy should be ignored. See [details below](#proxy-configuration). |
```

README.md
```diff
@@ -68,6 +69,12 @@ If your proxy server requires authentication embed the username and password in
 
 If your GitLab instance is exposed via plain HTTP (not recommended!) use `HTTP_PROXY` instead.
 
+#### Proxy bypass configuration
+
+You can configure a bypass option by setting `NO_PROXY` environment variable: `NO_PROXY=*.host.com, host.com`
+
+Depending on your gitlabUrl and a matching `NO_PROXY`it is bypassing your proxy configuration of the `HTTP_PROXY` or `HTTPS_PROXY` environment variable.
+
```
fgreinacher, line 77: Less is more :)

```suggestion
If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environment variable: NO_PROXY=*.host.com, host.com`

```

fgreinacher (changes requested):
Tanks @Babbafett, I left some comments.

Also I guess you want to format the code: `npm run lint -- --fix`

lib/resolve-config.js
Babbafett: Reworked on using proxy-from-env package

lib/resolve-config.js
Babbafett: Reworked on using proxy-from-env package

README.md
Babbafett: Added

README.md
Babbafett: Changed in last commit

README.md
```diff
@@ -69,11 +69,7 @@ If your proxy server requires authentication embed the username and password in
 
 If your GitLab instance is exposed via plain HTTP (not recommended!) use `HTTP_PROXY` instead.
 
-#### Proxy bypass configuration
-
-You can configure a bypass option by setting `NO_PROXY` environment variable: `NO_PROXY=*.host.com, host.com`
-
-Depending on your gitlabUrl and a matching `NO_PROXY`it is bypassing your proxy configuration of the `HTTP_PROXY` or `HTTPS_PROXY` environment variable.
+If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environment variable: NO_PROXY=*.host.com, host.com`
```
fgreinacher, line 72:
```suggestion
If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environment variable: `NO_PROXY=*.host.com, host.com`
```

lib/resolve-config.js
```diff
@@ -44,39 +43,41 @@ module.exports = (
...
-function getProxyConfiguration(gitlabUrl, HTTP_PROXY, HTTPS_PROXY) {
+function getProxyConfiguration(gitlabUrl) {
   const sharedParameters = {
     keepAlive: true,
     keepAliveMsecs: 1000,
     maxSockets: 256,
     maxFreeSockets: 256,
     scheduling: 'lifo',
   };
+  const proxy = getProxyForUrl(gitlabUrl);
```
fgreinacher, line 58: I don't really like that we are now using the environment variables implicitly. This is unexpected and reduced testability. Can you change that back, maybe just copying the relevant parts for `getProxyForUrl`?

fgreinacher (changes requested):

lib/resolve-config.js
Babbafett: I derived the relevant part then from the proxy-from-env package and changed it back so that no longer implicit environment variables are used.

lib/resolve-config.js
```diff
@@ -44,11 +45,69 @@ module.exports = (
     assets: assets ? castArray(assets) : assets,
     milestones: milestones ? castArray(milestones) : milestones,
     successComment,
-    proxy: getProxyConfiguration(defaultedGitlabUrl, HTTP_PROXY, HTTPS_PROXY),
+    proxy: getProxyConfiguration(defaultedGitlabUrl, HTTP_PROXY, HTTPS_PROXY, NO_PROXY),
   };
 };
 
-function getProxyConfiguration(gitlabUrl, HTTP_PROXY, HTTPS_PROXY) {
+function shouldProxy(gitlabUrl, NO_PROXY) {
```
fgreinacher, line 52: Not strictly required by the license, but still nice to give attribution:

```suggestion
// Copied from Rob Wu's great proxy-from-env library: https://github.com/Rob--W/proxy-from-env/blob/96d01f8fcfdccfb776735751132930bbf79c4a3a/index.js#L62
function shouldProxy(gitlabUrl, NO_PROXY) {

```

fgreinacher (approved):
Looking great now, thanks for your work @Babbafett!

Mitumaru (approved):

Issue #258, opened 2021-07-22
Documentation fix.


EvanCarroll:
I think it's a better to make these example uniform, it's really confusing.

`semantic-release/gitlab` (this project)
====

```json
{
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    ["@semantic-release/gitlab", {
      "gitlabUrl": "https://custom.gitlab.com",
      "assets": [
        {"path": "dist/asset.min.css", "label": "CSS distribution"},
        {"path": "dist/asset.min.js", "label": "JS distribution"}
      ]
    }],
  ]
}
```

1. It's not clear from the above how to publish an angular project, which builds into `dist` hand has many "assets". `npm pack` just needs to be told to the `dist/` directory that contains the `package.json`, for example `npm pack dist/my-workspace/my-lib`. How do we package an angular project with Semantic-Release gitlab like above?
2. You post a `gitlabUrl` in there. It's not clear why I would ever want that. That can be supported by the `package.json`. Why would I want it here. For simplicity I would remove it.

[GitLab wiki](https://docs.gitlab.com/ee/ci/examples/semantic-release.html),
====

The following is from the GitLab wiki,

```json
{
  "branches": ["master"],
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    "@semantic-release/gitlab",
    "@semantic-release/npm",
    [
      "@semantic-release/git",
      {
        "assets": ["package.json"],
        "message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
      }
    ]
  ]
}
```

1. Note this has `branches` That seems to be required. If it is this project should probably also contain it.

triwats:
This seems really sane to me and should be approved

JonasSchubert:
Hi. I created a small PR #366 to start working on this. Feel free to jump on and work on this. I just added the missing configuration for `branches`.

Regarding your remarks:
1. In your angular example: Do you want to create a NPM package or a zipped dist folder? If you want to create a NPM package hosted in the package registry, I think the documentation for the NPM plugin is more relevant.
2. I think the `gitlabUrl` is there to showcase some possibilities which will be explained further down. I agree that most people will not set this property because they will use the CI/CD. Here we use the provided env vars. But if one wants to use the CLI and has a private instance it might be useful to know.

These are just my thoughts. Feel free to add and change the mentioned PR to optimize the documentation 😉

PR #366, opened 2022-04-16
docs(usage): add branches to provide working example #258

JonasSchubert:
As highlighted in #258 the documentation for the usage might be a bit confusing. Without branches the configuration would not be valid. In my opinion we should provide a working example.


fgreinacher (approved):
Thanks!

PR #371, opened 2022-04-28
fix(fail): use encoded repo id to avoid error 404

JonasSchubert:
I checked the fail hook one more time and I found a bug. The encoded repo id should not be used in the issue api url. Otherwise Gitlab returns 404.


fgreinacher (approved):

fgreinacher:
Thanks!

Issue #382, opened 2022-05-24
Failing to publish assets with url without specifying `path`


Krewedk0:
Release 9.3.0 (#322) introduced a feature to publish assets with direct links when `url` field specified in asset.
But `path` field is still required for this to work, otherwise release fails with 

`The assets option (https://github.com/semantic-release/gitlab/blob/master/README.md#assets) must be an Array of Strings or Objects with a path property.`

This check wasn't changed here https://github.com/semantic-release/gitlab/blob/master/lib/verify.js#L18
as `path` should be optional in this case, according to the readme.

fgreinacher:
@awcjack Can you have a look?

awcjack:
Didn't realize it when rebase  
I will create pull request later

fgreinacher:
:tada: This issue is resolved in version 9.3.2 :tada:

The release is available on:
- [npm package (@latest dist-tag)](https://www.npmjs.com/package/@semantic-release/gitlab/v/9.3.2)
- [GitHub release](https://github.com/semantic-release/gitlab/releases/tag/v9.3.2)

Your **[semantic-release](https://github.com/semantic-release/semantic-release)** bot :package::rocket:

PR #387, opened 2022-05-29
fix(verify): Allow url asset not specifying path #382

awcjack:
Fix https://github.com/semantic-release/gitlab/issues/382

Validator accept url asset not specifying `path` property


lib/verify.js
```diff
@@ -15,7 +15,7 @@ const canBeDisabled = (validator) => (value) => value === false || validator(val
 
 const VALIDATORS = {
   assets: isArrayOf(
-    (asset) => isStringOrStringArray(asset) || (isPlainObject(asset) && isStringOrStringArray(asset.path))
+    (asset) => isStringOrStringArray(asset) || (isPlainObject(asset) && isNonEmptyString(asset.url)) || (isPlainObject(asset) && isStringOrStringArray(asset.path))
```
fgreinacher, line 18: What about adding some brackets for readability and getting rid of the double `isPlainObject` call?

```suggestion
    (asset) => isStringOrStringArray(asset) || (isPlainObject(asset) && (isNonEmptyString(asset.url) || isStringOrStringArray(asset.path)))
```

fgreinacher (changes requested):
Thanks @awcjack! 

I left a suggestion and we have some [lint issues](https://github.com/semantic-release/gitlab/runs/6644767911?check_suite_focus=true) 😄

lib/verify.js
awcjack: Done

fgreinacher (approved):

fgreinacher:
Thanks for taking care of this!

PR #390, opened 2022-06-02
docs: fix successComment documentation

milo526:
Two arguments were switched around and one of the links stated they were linking to GitHub while they were (and should be) linking to GitLab.


fgreinacher (approved):
Thanks for fixing this!

Issue #393, opened 2022-06-03
Failing (?) feature "allow adding URLs to release assets"


kaerbr:
## Current behavior

As of version [9.3.0](https://github.com/semantic-release/gitlab/releases/tag/v9.3.0) its possible to add urls to assets for linking to docker registry or other artifact storages.  
Unfortunately I cant get this feature to work for me.  
The URL to be used is given in my `.releaserc.yml`., but the pipeline fails (see at the bottom of this issue).

If I run the command manually (curl) (copied from the logs):
```
curl --request POST \
  --url https://gitlab.com/api/v4/projects/36721546/releases \
  --header 'Content-Type: application/json' \
  --header 'PRIVATE-TOKEN: <token>' \
  --data '{
	"tag_name": "v1.0.0",
	"description": "# 1.0.0 (2022-06-03)\n\n\n### Features\n\n* add url to release ([da40fc9](https://zzz/commit/da40c9e1606dc3e27f090ef13f51985bb))\n\n\n\n",
	"assets": {
		"links": [
			{
				"url": "https://gitlab.com/gitlab-org/gitlab/-/blob/master/README.md"
			}
		]
	}
}'
```
I get an error: `"error": "assets[links][0][name] is missing"` wich is understandable due to the fact that `assets:links:name` is required if a `assets:links:url` is given [(see here(docs))](https://docs.gitlab.com/ee/api/releases/#create-a-release)

## Expected behavior

Is there anything in my configuration not correct or is there a bug in this feature?  
I tried adding the `name` variable in my `.releaserc.yml` but it would not go into the request:
```
- - "@semantic-release/gitlab"
    - assets:
        - url: https://gitlab.com/gitlab-org/gitlab/-/blob/master/README.md
          name: "some name"
```

## Environment

- **semantic-release** version: 19.0.2
- CI environment: gitlab ci with `node:lts-alpine` base image
- Plugins used:  
  - npm install -g semantic-release@19 @semantic-release/changelog@6 @semantic-release/exec@6 @semantic-release/git@10 @semantic-release/gitlab@9 @semantic-release/npm@9 @saithodev/semantic-release-backmerge@2
- **semantic-release** configuration:  
```
plugins:
  - "@semantic-release/commit-analyzer"
  - "@semantic-release/release-notes-generator"
  - "@semantic-release/changelog"
  #- - "@semantic-release/exec"
[...]

PR #401, opened 2022-06-11
Retrieve link name from link when using link asset

awcjack:
link name is required for link asset following [documentation](https://docs.gitlab.com/ee/api/releases/#create-a-release)
If label property is not provided, try to retrieve the link name from url

Close https://github.com/semantic-release/gitlab/issues/393


lib/publish.js
```diff
@@ -48,7 +48,14 @@ module.exports = async (pluginConfig, context) => {
         const label = asset.label ? template(asset.label)(context) : undefined;
         const _url = asset.url;
         if (_url) {
-          assetsList.push({label, rawUrl: _url, type, filepath});
+          let _label = label;
+          if (!label) {
+            const parsedUrl = new URL(_url);
+            const file = pathlib.basename(parsedUrl.pathname);
+            _label = file ? file : 'Unknown asset';
```
fgreinacher, line 55: I don't like the fallback here. What about falling back to parsedUrl.hostname?

Also can any of these throw an exception that we should handle?

README.md
```diff
@@ -97,8 +97,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
...
 | -------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
 | `path`   | **Required**, unless `url` is set. A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
-| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images.                           | -                                    |
-| `label`  | Short description of the file displayed on the GitLab release. Can be dynamically adjusted with [Lodash template](https://lodash.com/docs#template). Allows same variables as [`successComment`](#successComment). Ignored if `path` matches more than one file. | File name extracted from the `path`. |
+| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. The url must be unique within the release.                           | -                                    |
```
fgreinacher, line 100:
```suggestion
| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. Must be unique within the release.                           | -                                    |
```

README.md
```diff
@@ -97,8 +97,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
...
+| `label`  | Short description of the file displayed on the GitLab release. Can be dynamically adjusted with [Lodash template](https://lodash.com/docs#template). Allows same variables as [`successComment`](#successComment). Ignored if `path` matches more than one file. The Label must be unique within the release when `url` is set. | File name extracted from the `path`. File name extracted from `url` |
```
fgreinacher, line 101: I'd simplify this a bit:

```suggestion
| `label`  | Short description of the file displayed on the GitLab release. Can be dynamically adjusted with [Lodash template](https://lodash.com/docs#template). Allows same variables as [`successComment`](#successComment). Ignored if `path` matches more than one file. Must be unique within the release. | File name extracted from `path` or `url` |
```

fgreinacher:
@awcjack Are you planning to continue here? If not just speak and I can finish it!

lib/publish.js
awcjack: You mean replacing 'Unknown asset' by `parsedUrl.hostname`?  
Actually I am thinking any better handling here too since the link name must be unique within the release.  

So when user have urls like below will cause exception during publish
https://server1.com/1.txt
https://server2.com/1.txt

JonasSchubert: We could include the index from `.map`, e.g.:
```js
_label = file ? file : `${parsedUrl.hostname}_${index}`;
```

PR #365, opened 2022-04-16
feat(assets): allow label templating #334

JonasSchubert:
As mentioned in #334 it could be useful to use templating for the asset labels. This PR provides the functionality to use the lodash templating functionality similar to the `successComment` and the `failComment`.

Resolves #334

PR #409, opened 2022-06-20
Allow templating

### LINKED ISSUE #406: asset url containing ${nextRelease.version} not working/expanded
I would like to create version dependent `url`s as release assets pointing to the specific file on an other filehosting server.

```json
      "assets": [
          {"url": "http://fileserver/${nextRelease.version}/update.zip", "label": "update file", "type": "other"}
      ]
}]
```

The release page nevertheless will contain the following link:
```
http://fileserver/$%7BnextRelease.version%7D/update.zip
```

For the asset label that seems to work as the [documentation suggests](https://github.com/semantic-release/gitlab#usage), e.g.
```json
{"path": "dist/asset.min.js", "label": "v${nextRelease.version}.js"}
```

But why not for the `url`?

kaerbr:
I added templating functioniality to `url`, `type` and ` filepath` like already done in #365 for `label`.
Closes #406 

Hope you like it (:

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` 
 | `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. Supports [Lodash templating](https://lodash.com/docs#template).                           | -                                    |
 | `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file. Supports [Lodash templating](https://lodash.com/docs#template). | File name extracted from the `path`. |
 | `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). Supports [Lodash templating](https://lodash.com/docs#template). | `other` |
-| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. | - |
+| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. Supports [Lodash templating](https://lodash.com/docs#template). | - |
 
 Each entry in the `assets` `Array` is globbed individually. A [glob](https://github.com/isaacs/node-glob#glob-primer)
 can be a `String` (`"dist/**/*.js"` or `"dist/mylib.js"`) or an `Array` of `String`s that will be globbed together
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` 
 | `path`   | **Required**, unless `url` is set. A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
 | `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. Supports [Lodash templating](https://lodash.com/docs#template).                           | -                                    |
 | `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file. Supports [Lodash templating](https://lodash.com/docs#template). | File name extracted from the `path`. |
-| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). | `other` |
+| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). Supports [Lodash templating](https://lodash.com/docs#template). | `other` |
 | `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. | - |
 
 Each entry in the `assets` `Array` is globbed individually. A [glob](https://github.com/isaacs/node-glob#glob-primer)
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array` 
 | Property | Description                                                                                                 | Default                              |
 | -------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
 | `path`   | **Required**, unless `url` is set. A [glob](https://github.com/isaacs/node-glob#glob-primer) to identify the files to upload.    | -                                    |
-| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images.                           | -                                    |
-| `label`  | Short description of the file displayed on the GitLab release. Can be dynamically adjusted with [Lodash template](https://lodash.com/docs#template). Allows same variables as [`successComment`](#successComment). Ignored if `path` matches more than one file. | File name extracted from the `path`. |
+| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. Supports [Lodash templating](https://lodash.com/docs#template).                           | -                                    |
+| `label`  | Short description of the file displayed on the GitLab release. Ignored if `path` matches more than one file. Supports [Lodash templating](https://lodash.com/docs#template). | File name extracted from the `path`. |
 | `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). | `other` |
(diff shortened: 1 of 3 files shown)


MrSplint3r:
Why did you do these changes that can be seen here?
https://github.com/semantic-release/gitlab/compare/cc91eddc3bc7bbe7eee1253baeb2f51f1241d7d1..65deb545bccef9b779c84df87741c6af9e2f2ce8

This means you are using now template literals on your test, which is replacing the `${...}` expressions before Lodash can even get to it. The $ needs to be kept escaped for them to be replaced only at Lodash time. Or simply don't use the backticks.

Your tests where you try to use `process.env.XXX` vars will most likely not work since the process object is not present in the context given to the Lodash template here: https://github.com/semantic-release/gitlab/blob/65deb545bccef9b779c84df87741c6af9e2f2ce8/lib/publish.js#L48

kaerbr:
> Why did you do these changes that can be seen here? https://github.com/semantic-release/gitlab/compare/cc91eddc3bc7bbe7eee1253baeb2f51f1241d7d1..65deb545bccef9b779c84df87741c6af9e2f2ce8
> 
> This means you are using now template literals on your test, which is replacing the `${...}` expressions before Lodash can even get to it. The $ needs to be kept escaped for them to be replaced only at Lodash time. Or simply don't use the backticks.

The linting stage of the test was failing before this change. But you are right that lodash wont get to the `${...}` expressions. I will add a change. (:

> Your tests where you try to use `process.env.XXX` vars will most likely not work since the process object is not present in the context given to the Lodash template here:
> 
> https://github.com/semantic-release/gitlab/blob/65deb545bccef9b779c84df87741c6af9e2f2ce8/lib/publish.js#L48

But why would the tests run if this was the case? -- sry I'm new to Javascript. :(
Ok so I did a testrun with semantic-release/gitlab@9.3.2 (since [v9.2.0](https://github.com/semantic-release/gitlab/releases/tag/v9.2.0) label templating is implemented) :
- relevant parts of the .releaserc.yml:
```
- - "@semantic-release/gitlab"
    - assets:
        - url: "https://my.url"
          label: "${process.env.LABEL}"
          type: other
```
- relevant parts of the .gitlab-ci.yml:
```
semantic-release:
  stage: release
  script:
    - export LABEL="my-label"
    - npm install -g semantic-release@19 @semantic-release/changelog@6 @semantic-release/exec@6 @semantic-release/git@10 @semantic-release/gitlab@9
    - semantic-release --debug
```
- relevant parts of the debug output:
```
2022-06-22T06:02:52.221Z semantic-release:gitlab POST-ing the following JSON to https://gitlab.com/api/v4/projects/kaerbr%2Fsemantic-release-test/releases:
{
  "tag_name": "v1.8.0",
  "description": "# [1.8.0](https://gitlab.com/kaerbr/semantic-release-test/compare/v1.7.0...v1.8.0) (2022-06-22)\n\n\n### Features\n\n* process:env.XXX ([a361fd5](https://gitlab.com/kaerbr/semantic-release-test/commit/a361fd5bce4785ffeb9c8f6c266507700ceefa00))\n\n\n\n",
  "assets": {
    "links": [
      {
        "name": "my-label",
        "url": "https://my.url/",
        "link_type": "other"
      }
    ]
  }
}
```

So since I implemented the changes like for the label templating I think this should work and be fine? (:

MrSplint3r:
Sorry, I was mistaken, the env vars object is indeed available on the context passed to the template functions, just under the name `env`. See here:
https://github.com/semantic-release/semantic-release/blob/9589a96239826abe9b07e8deffcc7d8aeb9c2e40/index.js#L257
https://github.com/semantic-release/semantic-release/blob/9589a96239826abe9b07e8deffcc7d8aeb9c2e40/index.js#L250

In that sense, now I understood why your tests are succeeding, even though you were using `${process.env.XXX}` instead of `${env.XXX}`. That's because Lodash also interprets that as a JavaScript expression and since `process.env` is available globally, it was able to resolve it.

I could definitely use this functionality as well so from me a 👍  for the contribution. :)

test/publish.test.js
```diff
@@ -176,6 +176,133 @@ test.serial('Publish a release with an asset with a template label', async (t) =
...
+
+  const result = await publish({assets}, {env, cwd, options, nextRelease, logger: t.context.logger});
+
+  t.is(result.url, `https://gitlab.com/${owner}/${repo}/-/releases/${encodedGitTag}`);
+  t.deepEqual(t.context.log.args[0], ['Uploaded file: %s', uploaded.url]);
+  t.deepEqual(t.context.log.args[1], ['Published GitLab release: %s', nextRelease.gitTag]);
+  t.true(gitlabUpload.isDone());
+  t.true(gitlab.isDone());
+});
+
```
fgreinacher, line 305: These tests are super-similar, what about just having one that we parameterize?

README.md
```diff
@@ -97,10 +97,10 @@ Can be a [glob](https://github.com/isaacs/node-glob#glob-primer) or and `Array`
...
-| `label`  | Short description of the file displayed on the GitLab release. Can be dynamically adjusted with [Lodash template](https://lodash.com/docs#template). Allows same variables as [`successComment`](#successComment). Ignored if `path` matches more than one file. | File name extracted from the `path`. |
-| `type` | Asset type displayed on the GitLab release. Can be `runbook`, `package`, `image` and `other` (see official documents on [release assets](https://docs.gitlab.com/ee/user/project/releases/#release-assets)). | `other` |
-| `filepath` | A filepath for creating a permalink pointing to the asset (requires GitLab 12.9+, see official documents on [permanent links](https://docs.gitlab.com/ee/user/project/releases/#permanent-links-to-release-assets)). Ignored if `path` matches more than one file. | - |
+| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. Can be dynamically adjusted with [Lodash template](https://lodash.com/docs#template). Allows same variables as [`successComment`](#successComment) + all given environment variables.                           | -                                    |
```
fgreinacher, line 100: I don't like that we duplicate this text everywhere. What about adding a new section `Templating` and just have something like this here:

```suggestion
| `url`    | Alternative to setting `path` this provides the ability to add links to releases, e.g. URLs to container images. Supports [templating](#templating). | -                                    |
```

fgreinacher:
Thanks for your work @kaerbr, I left some comments, ping me if you need support!

test/publish.test.js
```diff
@@ -176,6 +176,133 @@ test.serial('Publish a release with an asset with a template label', async (t) =
...
+  const result = await publish({assets}, {env, cwd, options, nextRelease, logger: t.context.logger});
+
+  t.is(result.url, `https://gitlab.com/${owner}/${repo}/-/releases/${encodedGitTag}`);
+  t.deepEqual(t.context.log.args[0], ['Published GitLab release: %s', nextRelease.gitTag]);
+  t.true(gitlab.isDone());
+});
+
+test.serial('Publish a release (with an link) with a template type', async (t) => {
+  process.env.TYPE = 'other';
```
fgreinacher, line 219: Should we reset this at some point so that it does not interfere with other tests?

kaerbr: Just now, I dont think there will be any interference, but better safe than sorry!👍🏼

test/publish.test.js
kaerbr: Do you have any example how to make a parameterized test in JS?
I only have experience with Java for this scenario.

fgreinacher: Ava, our test runner, [does not have a built-in way to do this](https://github.com/avajs/ava/issues/2414). Either wrap the entire `test.serial(...)` in a loop or add [`pava`](https://github.com/TomerAberbach/pava)

kaerbr: What do you think about putting all in one test? (: -- I mean its nearly the same and for me more like an integration rather than an unit test.

```
test.serial('Publish a release (with an link) with variables', async (t) => {
  process.env.TYPE = 'other';
  const cwd = 'test/fixtures/files';
  const owner = 'test_user';
  const repo = 'test_repo';
  const env = {GITLAB_TOKEN: 'gitlab_token'};
  const nextRelease = {gitHead: '123', gitTag: 'v1.0.0', notes: 'Test release note body', version: '1.0.0'};
  const options = {repositoryUrl: `https://gitlab.com/${owner}/${repo}.git`};
  const encodedRepoId = encodeURIComponent(`${owner}/${repo}`);
  const encodedGitTag = encodeURIComponent(nextRelease.gitTag);
  const assets = [
    {
      label: 'README-v\${nextRelease.version}.md',
      type: `\${process.env.TYPE}`,
      url: `https://gitlab.com/gitlab-org/gitlab/-/blob/master/README-v\${nextRelease.version}.md`,
    },
  ];
  const gitlab = authenticate(env)
    .post(`/projects/${encodedRepoId}/releases`, {
      tag_name: nextRelease.gitTag,
      description: nextRelease.notes,
      assets: {
        links: [
          {
            name: 'README-v1.0.0.md',
            url: `https://gitlab.com/gitlab-org/gitlab/-/blob/master/README-v1.0.0.md`,
            link_type: 'other',
          },
        ],
      },
    })
    .reply(200);

  const result = await publish({assets}, {env, cwd, options, nextRelease, logger: t.context.logger});

  t.is(result.url, `https://gitlab.com/${owner}/${repo}/-/releases/${encodedGitTag}`);
  t.deepEqual(t.context.log.args[0], ['Published GitLab release: %s', nextRelease.gitTag]);
  t.true(gitlab.isDone());
});
```

fgreinacher:
@kaerbr Thanks for adapting the tests! Ping me once you addressed the other comments - also if I should take over something!

test/publish.test.js
kaerbr: I reset the variables in line 237 & 238:
```
delete process.env.TYPE;
delete process.env.FILEPATH;
```

README.md
kaerbr: @fgreinacher I added a link to the lodash docs (template section) (:

fgreinacher (approved):
Thanks @kaerbr, looking great!

PR #440, opened 2022-08-31
feat(publish): add option to skip publishing releases

redabacha:
this adds a simple flag to skip the publish step which is useful for when you have development packages that don't need a gitlab release but would benefit from having the success comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ If you need to bypass the proxy for some hosts, configure the `NO_PROXY` environ
 | `failTitle`           | The title of the issue created when a release fails. Set to `false` to disable opening an issue when a release fails.                                           | `The automated release is failing 🚨`                                                                                                                                    |
 | `labels`              | The [labels](https://docs.gitlab.com/ee/user/project/labels.html#labels) to add to the issue created when a release fails. Set to `false` to not add any label. Labels should be comma-separated as described in the [official docs](https://docs.gitlab.com/ee/api/issues.html#new-issue), e.g. `"semantic-release,bot"`. | `semantic-release`                                                                                                                                                      |
 | `assignee`            | The [assignee](https://docs.gitlab.com/ee/user/project/issues/managing_issues.html#assignee) to add to the issue created when a release fails.                  | -                                                                                                                                                                       |
+| `skipPublish` | Skips the publish step so a GitLab release will not be created. This is useful if you only want to add comments for example. | `false` |
 
 #### assets
 
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -18,7 +18,16 @@ module.exports = async (pluginConfig, context) => {
     nextRelease: {gitTag, gitHead, notes},
     logger,
   } = context;
-  const {gitlabToken, gitlabUrl, gitlabApiUrl, assets, milestones, proxy} = resolveConfig(pluginConfig, context);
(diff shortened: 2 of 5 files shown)


fgreinacher:
We currently don't have this feature for the GitHub plugin and we generally try to keep both plugins as similar as possible. What do you think about this @gr2m @travi?

travi:
this feels very much like an edge case to me. what is the harm of having the release created, even if your team is not using it? we are trying to keep complexity down in order to avoid unnecessary maintenance that we need to carry forward, so we prefer to focus on options that solve impactful issues or are needed by many users. this doesnt feel like either of those to me, but i'm open to being convinced otherwise.

side note for the future, it is recommended to start with an issue or discussion when proposing new functionality to ensure agreement in the value so that you can avoid investment of effort into a PR if there isnt agreement that the proposal should move forward.

redabacha:
> what is the harm of having the release created, even if your team is not using it?

it creates a huge amount of spam releases if we make a lot of commits which we typically do which makes it nearly impossible to view stable releases of our package in the releases page on gitlab due to the clutter. 

> solve impactful issues or are needed by many users

the ability to have comments added by semantic-release on gitlab is a fairly recent addition and this is not a complicated option to have nor maintain imo. 

> you can avoid investment of effort into a PR

i already had a patch for this package pretty much identical to what this pr does so there wasn't much that had to be done except the readme changes and the unit test. i believe my use case is valid and i don't see any alternative solution to get this behavior with the existing options provided by this package?

travi:
> view all stable releases of our package

since you differentiate stable releases, are you implying that the desire for this is only for pre-releases? your description above does not mention that and the config change in this contribution would not differentiate stable vs not. 

could you please clarify what you mean by "development packages" vs packages where you _would_ want gitlab releases? also, please further clarify what you mean by stable releases vs the ones you want this to apply to. 

it may also be helpful to understand what your workflow looks like. are you using git flow, github flow, trunk-based development, or something else?

travi:
ok, thats what i was starting to gather based on the details that you had provided. unfortunately, using pre-releases to simulate git-flow is not a supported workflow and is not one that we want to add new features in order to support. 

semantic-release is not a general purpose workflow tool and the decision was made long ago to avoid supporting git-flow. we encourage use of github flow or trunk-based development for your packages, even if you use git-flow for the applications that consume your packages. 

you are welcome to maintain your own fork if you decide this is behavior you want to use with your projects, but we do not intend to improve support for git-flow and cannot promise that future changes will not make usage with git-flow more difficult.

fgreinacher:
Closing this as per the comments above.

Thanks a lot for your willingness to contribute back though 👍

Guilhermeasper:
I was about to open an issue related to this matter when I came across this closed PR. Would you consider adding a `skipPrerelease` option to prevent the publish of releases on GitLab when dealing with a branch configured as a prerelease?

Currently, I’m facing the same issue. My team makes frequent changes to the develop branch, which is set as a prerelease. As a result, I end up with numerous prerelease entries cluttering the releases page, making it challenging to find stable releases.

I’m asking if you would consider this enhancement because I could open a PR for both packages (GitLab and GitHub) since you aim to keep them in sync.

_p.s.: As a workaround, I currently have a custom plugin that sends a request to remove prereleases from the releases during the success step._

travi:
Please read the conversation above. Git flow is not a supported workflow and we are not interested in accepting contributions to support it

PR #464, opened 2022-12-08
feat: convert package to ESM

fgreinacher:
Contributes to https://github.com/semantic-release/semantic-release/issues/2133


.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           cache: npm
-          node-version: 16
+          node-version: 18
```
travi, line 18: we just ran https://github.com/gr2m/octoherd-script-bump-node-version-in-workflows/ against the org to use `lts/*` here so that we dont have to keep updating these over time. i'd pull that in here in place of this change

travi (commented):
looks great!

what would you think about releasing this as a beta before promoting to stable? we just released a beta of the npm plugin so we could try it in some real releases first since we havent fully proven out that core is ready to load esm plugins quite yet

travi:
> what would you think about releasing this as a beta before promoting to stable? we just released a beta of the npm plugin so we could try it in some real releases first since we havent fully proven out that core is ready to load esm plugins quite yet

in our [attempt to release a beta of core](https://github.com/semantic-release/semantic-release/pull/2670) that included the beta of the npm plugin, we confirmed that we needed another iteration to handle loading esm plugins properly. until we promote that fix to stable, i think i'd advise against releasing this as stable quite yet

.github/workflows/release.yml
fgreinacher: Cool, thanks for the hint! Adopted changes from `main`.

fgreinacher:
> what would you think about releasing this as a beta before promoting to stable? we just released a beta of the npm plugin so we could try it in some real releases first since we havent fully proven out that core is ready to load esm plugins quite yet

Good idea, I rebased this against a `beta` branch. This should be all that's necessary to publish a beta release, right?

travi (approved):

travi:
> I rebased this against a `beta` branch. This should be all that's necessary to publish a beta release, right?

Yep that'll do it

travi:
🎉 great to see this make it's way out. Thanks for moving this forward, @fgreinacher!

One thing that I would highlight is that the esm transition should be considered breaking, but looks like it wasn't released that way in this case. Could you re-release to call out that breaking change? That way when promoted to stable the release will cover it then too. Also, when we do promote to stable, I recommend a normal merge so the releases from the pre-release branch are preserved.

fgreinacher:
Thanks for your thoughts @travi! Yea, I noticed that I forgot the breaking change trailer. What’s the best way to perform a re-release? Delete the tag and force-push?

Regarding the merge strategy: Currently this repo is configured to enforce squash merge. Could you change that setting?

travi:
> What’s the best way to perform a re-release? Delete the tag and force-push?

yeah, all options are a bit ugly, but i that approach should work fine

> Currently this repo is configured to enforce squash merge. Could you change that setting?

i updated to enable squash or normal merges, so you should be good to go once you're ready to promote to stable. just a reminder that you'll likely want to wait until one of the options [mentioned here](https://github.com/semantic-release/semantic-release/pull/2670#discussion_r1070214972) (promoting the current core beta to stable, or backporting the handling of esm plugins to core) make their way to the stable version of core.

PR #483, opened 2023-01-18
Promote changes from beta to main

fgreinacher:
🛑 Don't squash when merging


travi:
support for loading ESM plugins has been backported to the current stable channel in https://github.com/semantic-release/semantic-release/releases/tag/v20.1.0. i'd suggest trying out your beta with that version, but without any surprises, i expect that blocker to be solved.

package.json
```diff
@@ -70,14 +71,12 @@
...
-  },
   "prettier": {
     "printWidth": 120,
-    "trailingComma": "es5",
-    "singleQuote": true,
-    "bracketSpacing": false
+    "trailingComma": "es5"
+  },
+  "peerDependencies": {
+    "semantic-release": ">=18.0.0"
```
travi, line 79: since this version of the gitlab plugin will require a version of semantic-release that is capable of loading ESM plugins, i would recommend setting this to `>=v20.1.0`

fgreinacher:
> support for loading ESM plugins has been backported to the current stable channel in https://github.com/semantic-release/semantic-release/releases/tag/v20.1.0. i'd suggest trying out your beta with that version, but without any surprises, i expect that blocker to be solved.

Yes, works perfectly with `semantic-release@20.1.0` and the `@semantic-release/gitlab@10.0.0-beta.2`: https://gitlab.com/fgreinacher/semantic-release-test/-/jobs/3666245807

travi (approved):
🎉 🚀

PR #487, opened 2023-01-29
fix: move prettier to devDependencies

mat813:
It is not used anywhere in the code and is only needed during development.

Also, some people, like me, depend on a [fork of prettier](https://www.npmjs.com/package/@btmills/prettier) and now, depending on the order the dependencies get installed, either the fork or the original get used, which makes ci/cd a nightmare because builds gets unreproducible.


fgreinacher (approved):
Good catch, thanks!

PR #491, opened 2023-02-01
fix(generic-package): fix issues when uploading multiple assets

### LINKED ISSUE #490: Generic package from registry can't be extracted
Hi everyone 👋 

I'm using GitLab registry to push an asset with the format `gzip compressed data` (.tar.gz) but when I download the uploaded package, I can't decompress it.

Follow my `.releaserc.yml`

```yaml
branches:
  - "main"
plugins:
  - "@semantic-release/commit-analyzer"
  - "@semantic-release/release-notes-generator"
  - - "@semantic-release/exec"
    - verifyConditionsCmd: 'tar -czf package.tar.gz --exclude=".git*" --exclude="__pycache__" aws/'
  - - "@semantic-release/gitlab"
    - assets:
        - path: "package.tar.gz"
          filepath: "/package.tar.gz"
          label: "package.tar.gz"
          target: generic_package
          status: default
tagFormat: ${version}
```

If we do a `cat` to the package file, we will see weird data on top and at the end of the file.

```bash
❯ cat package.tar.gz
----------------------------636039438515691214353669
Content-Disposition: form-data; name="file"; filename="aws.tar.gz"
Content-Type: application/gzip

<REDACTED BINARY DATA>
----------------------------636039438515691214353669--
```

If we remove the first 4 files from the file with the command `sed -e '1,4d' < package.tar.gz > package2.tar.gz`, we will decompress it.

If we compress the file manually, upload it, and download it after we can decompress it since the data are been untouched.

Can someone give any insight?

JonasSchubert:
This should fix #490 (WIP). Working on it I found some issues which were not recognized yet:
- wrong README example
- missing proxy
- invalid URL if `gitlapApiUrl` is set

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The plugin can be configured in the [**semantic-release** configuration file](ht
         "gitlabUrl": "https://custom.gitlab.com",
         "assets": [
           { "path": "dist/asset.min.css", "label": "CSS distribution" },
-          { "path": "dist/asset.min.js", "label": "JS distribution", "type": "generic_package" },
+          { "path": "dist/asset.min.js", "label": "JS distribution", "target": "generic_package" },
           { "path": "dist/asset.min.js", "label": "v${nextRelease.version}.js" },
           { "url": "https://gitlab.com/gitlab-org/gitlab/-/blob/master/README.md" }
         ]
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -149,7 +149,7 @@ export default async (pluginConfig, context) => {
       links: assetsList.map(({ label, alt, url, type, filepath, rawUrl, target }) => {
         return {
           name: label || alt,
-          url: rawUrl || (target === 'generic_package' ? url : urlJoin(gitlabUrl, repoId, url)),
+          url: rawUrl || (target === "generic_package" ? url : urlJoin(gitlabUrl, repoId, url)),
           link_type: type,
           filepath,
         };
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -1,4 +1,4 @@
-import { createReadStream } from "fs";
+import { createReadStream, readFileSync } from "fs";
 import pathlib from "path";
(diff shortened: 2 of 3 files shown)


JonasSchubert:
Now I am experiencing issues with `got` stream handling. When trying to upload more then one file as asset, I receive this error: `An error occurred while running semantic-release: HTTPError: Cannot retry with consumed body stream`

The first file is successfully uploaded, after this everything fails.

@fgreinacher I am not sure how to handle this using `got` and would appreciate your input (or anybodys input regarding this :))

Config:
```json
{
  "branches": ["feat/text-semantic-release-gitlab-assets"],
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    [
      "@semantic-release/gitlab",
      {
        "gitlabUrl": "https://code.siemens.com",
        "assets": [
          { "path": "generic-package.tar", "label": "tar-example.tar", "target": "generic_package", "status": "default" },
          { "path": "generic-package.7z", "label": "7z-example.7z", "target": "generic_package" },
          { "path": "generic-package.txt", "label": "txt-example.txt", "target": "generic_package", "status": "hidden" }
        ]
      }
    ]
  ]
}
```

The file is put into a tar and 7z using 7zip:
[generic-package.txt](https://github.com/semantic-release/gitlab/files/10556169/generic-package.txt)

JonasSchubert:
I am using `readFileSync` for the generic package. This resolved the issue with `got` streams. Or has there been a reason to use file streams?

fgreinacher:
The changes look good to me. Do you want to make additional changes @JonasSchubert?

JonasSchubert:
Thanks for the feedback @fgreinacher.
I would say let's merge it. If there will be a issue considering big files I would come back and help there, but in my opinion git LFS could be a better solution then - depending on the use case.

fgreinacher (approved):
Thanks for taking care!

PR #503, opened 2023-02-16
fix(generic-packages): use version for publish URL for consistency with npm

JonasSchubert:


fgreinacher:
LGTM, thanks!

I tend to mark this as a breaking change. What do you think about it @JonasSchubert?

JonasSchubert:
Sounds reasonable 👍
Even so it will not change anything with existing releases, future release will have slightly different URLs.
Most users will not feel any change, but e.g. automated scripts will have issues if not updated to the new URLs.

fgreinacher:
That were also my thoughts!

Will you add the breaking change trailer to the MR description?

JonasSchubert:
@fgreinacher @mfoti I would like to avoid two breaking changes shortly after each other as we already had one with the `got` upgrade.

So I propose to release this PR and https://github.com/semantic-release/gitlab/pull/504 by @mfoti together, if we agree on it. Either you (@mfoti) can take over my changes or I add yours here. I am fine either way.

#504 is related to the latest discussion in #343. If you don't mind I would like to hear your thoughts regarding this @fgreinacher

fgreinacher:
> @fgreinacher @mfoti I would like to avoid two breaking changes shortly after each other as we already had one with the `got` upgrade.
> 
> So I propose to release this PR and #504 by @mfoti together, if we agree on it. Either you (@mfoti) can take over my changes or I add yours here. I am fine either way.

Yes, this makes sense. You can keep the commits separate. The release is generated from the merge commit.

I'm on vacation until Wednesday. Just ping me once you are ready and I can merge.

JonasSchubert:
@fgreinacher @mfoti I cherry-picked the changes and adapted to my suggestions in the [PR](https://github.com/semantic-release/gitlab/pull/504).

- I added links to the Gitlab docs to quick link why we made the changes - buy we can remove them if not needed
- maybe the commit message can be optimised

fgreinacher (approved):
LGTM, thanks @JonasSchubert and @mfoti!

Issue #543, opened 2023-05-26
Use CI_PROJECT_ID instead of name of the project in API calls


gaborauth:
Hi,

I'm trying to use the plugin in my environment and project structure and I feel like I need either some support or I have an idea for a new feature... :)

So, I have some project with group/subgroup/project structure, for example `others/toolsforboox/toolsboox-backend`.

I have a minimalistic `.releaserc.yml`:
```
branches: ['main']

plugins:
  - "@semantic-release/changelog"
  - "@semantic-release/gitlab"
  - "@semantic-release/git"
```

The `semantic-release` detects the URLs via the environment variables:
```
$ echo $CI_PROJECT_PATH
others/toolsforboox/toolsboox-backend
$ echo $CI_PROJECT_URL
https://my.gitlab.url/others/toolsforboox/toolsboox-backend
$ echo $CI_PROJECT_ID
12
```

Based on the environment variables, the `semantic-release` try to act via the Gitlab API.
```
[7:15:55 AM] [semantic-release] › ✔  Run automated release from branch main on repository https://my.gitlab.url/others/toolsforboox/toolsboox-backend.git
[7:15:56 AM] [semantic-release] › ✔  Allowed to push to the Git repository
[7:15:56 AM] [semantic-release] › ℹ  Start step "verifyConditions" of plugin "@semantic-release/changelog"
[7:15:56 AM] [semantic-release] › ✔  Completed step "verifyConditions" of plugin "@semantic-release/changelog"
[7:15:56 AM] [semantic-release] › ℹ  Start step "verifyConditions" of plugin "@semantic-release/gitlab"
2023-05-26T07:15:56.588Z semantic-release:gitlab apiUrl: 'https://my.gitlab.url/api/v4'
2023-05-26T07:15:56.588Z semantic-release:gitlab repoId: 'others/toolsforboox/toolsboox-backend
```

But the API call based on the `repoId` fails:
```
[7:15:57 AM] [semantic-release] › ✘  EMISSINGREPO The repository others/toolsforboox/toolsboox-backend doesn't exist.
```

When I tried to retrieve the project based on the `repoId`, like the plugin, it fails:
```
$ curl --header "PRIVATE-TOKEN: access-token" https://my.gitlab.url/api/v4/projects/others/toolsforboox/toolsboox-backend
{"error":"404 Not Found"}
```

But, based on the numeric project ID, it works:
```
[...]

PR #576, opened 2023-08-03
fix: use numeric project ID for API calls

fgreinacher:
Before we used the project path, which can cause issues with some proxy or firewall setups that replace slashes in URLs.

Closes https://github.com/semantic-release/gitlab/issues/543

### DIFF
--- a/lib/get-repo-id.js
+++ b/lib/get-repo-id.js
@@ -1,9 +1,9 @@
 import parseUrl from "parse-url";
 import escapeStringRegexp from "escape-string-regexp";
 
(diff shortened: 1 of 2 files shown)


gr2m (approved):

JedaiRussia:
This fix caused the downloaded files to be unavailable. 
The assets files url now contains the Project ID instead of the project name, which made the files inaccessible. Manually fixing the link means that the file is available for download. 
Perhaps this is our problem with self-hosted GitLab. We decided to roll back to semantic-release/gitlab ver 12.0.3

PR #580, opened 2023-08-09
fix: use project path for API calls again

### LINKED ISSUE #579: Release asset URLs are incorrect
> This fix caused the downloaded files to be unavailable. 
> The assets files url now contains the Project ID instead of the project name, which made the files inaccessible. Manually fixing the link means that the file is available for download. 
> Perhaps this is our problem with self-hosted GitLab. We decided to roll back to semantic-release/gitlab ver 12.0.3
> 

_Originally posted by @JedaiRussia in https://github.com/semantic-release/gitlab/issues/576#issuecomment-1670059770_

fgreinacher:
The numeric project ID is not working for all cases and I don't have the capacity right now to think about a generic solution.

People that have issues with the project path can override `CI_PROJECT_PATH` for the semantic-release invocation. 

Reverts semantic-release/gitlab#576

Closes #579 

/cc @JedaiRussia @gaborauth

### DIFF
--- a/lib/get-repo-id.js
+++ b/lib/get-repo-id.js
@@ -1,9 +1,9 @@
 import parseUrl from "parse-url";
 import escapeStringRegexp from "escape-string-regexp";
 
-export default ({ envCi: { service } = {}, env: { CI_PROJECT_ID } }, gitlabUrl, repositoryUrl) =>
-  service === "gitlab" && CI_PROJECT_ID
-    ? CI_PROJECT_ID
+export default ({ envCi: { service } = {}, env: { CI_PROJECT_PATH } }, gitlabUrl, repositoryUrl) =>
+  service === "gitlab" && CI_PROJECT_PATH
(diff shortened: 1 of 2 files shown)


gaborauth:
I'm sad to hear that, I'm trying to make a private fork for ourself that works for us.

Anyway, thanks for trying (I'm currently on vacation, so I couldn't try it).

gr2m (approved):

glanceon:
> I'm sad to hear that, I'm trying to make a private fork for ourself that works for us.
> 
> Anyway, thanks for trying (I'm currently on vacation, so I couldn't try it).

Stumbled upon same issue, any updates ?

gaborauth:
> Stumbled upon same issue, any updates ?

It's released and reverted as I know. So, I created a fork, added a configuration option to use the CI_PROJECT_ID and use it internally.

PR #647, opened 2023-12-11
chore: bump minimum Node version to 20.8.1

fgreinacher:
This bumps the minimum supported Node version to 20.8.1. Node 20 is the new LTS version and some of our dependencies already dropped supported for Node 18. Check out [our Node Support Policy](https://semantic-release.gitbook.io/semantic-release/support/node-support-policy) for how we handle Node versions.

BREAKING CHANGE: Node 18 and 19 are not supported anymore. Users have to upgrade to Node 20.8.1 or later.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,8 +13,9 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 18.0.0
-          - 19
+          - 20.8.1 # minimum supported Node version
+          - 20
+          - 21
         os:
           - ubuntu-latest
           - macos-latest
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "debug": "^4.0.0",
     "dir-glob": "^3.0.0",
     "escape-string-regexp": "^5.0.0",
-    "form-data": "^4.0.0",
+    "formdata-node": "^6.0.3",
     "fs-extra": "^11.0.0",
     "globby": "^14.0.0",
     "got": "^13.0.0",
@@ -41,7 +41,7 @@
     "tempy": "1.0.1"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20.8.1"
   },
   "files": [
     "lib",
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -1,8 +1,9 @@
-import { createReadStream, readFileSync } from "fs";
+import { readFileSync } from "fs";
 import pathlib from "path";
 import fs from "fs-extra";
 import { isPlainObject, template } from "lodash-es";
-import FormData from "form-data";
+import { FormData } from "formdata-node";
+import { fileFromPath } from "formdata-node/file-from-path";
 import urlJoin from "url-join";
 import got from "got";
 import _debug from "debug";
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "tempy": "1.0.1"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20.8.1"
```
fgreinacher, line 44: I picked this because 20.8.1 is the most recent security update

fgreinacher:
@gr2m @travi You might have a better overview than me here. Have you ever seen issues like the one below? If not I'll start investigating.

```
  publish › Publish a release (with an link) with variables

  Error: Promise returned by test never resolved

  Error: Promise returned by test never resolved
      at Test.finishDueToInactivity (file:///home/runner/work/gitlab/gitlab/node_modules/ava/lib/test.js:587:9)
      at Runner.beforeExitHandler (file:///home/runner/work/gitlab/gitlab/node_modules/ava/lib/runner.js:241:13)
      at process.emit (node:events:514:28)
```

package.json
travi: makes sense to me. also aligns with recently restricted ranges i've seen:
* [ava](https://github.com/avajs/ava/blob/5975b602b771e0dc02382d24b65c5561bd5fc7ee/package.json#L39)
* [got](https://github.com/sindresorhus/got/releases/tag/v14.0.0)

travi:
> Have you ever seen issues like the one below?

only when the code under test is missing an await when needed. since you arent changing any tests with this, i assume it has to be something related to the node versions now being tested against. 

i've run into some situations related to changes with `node:fs`, like https://github.com/tschaub/mock-fs/issues/380. i wonder if `tempy` maybe has any similar situations. with that particular change, could try before and after node v20.8 to see if there are any differences in the success or failure.

fgreinacher:
I debug this a bit. All failures occur because the POST request to upload the files does not resolve:

https://github.com/semantic-release/gitlab/blob/master/lib/publish.js#L127

I noticed that our `FormData` library (https://www.npmjs.com/package/form-data) is a quite dated and have played around a bit with other libraries (https://www.npmjs.com/package/formdata-node and https://www.npmjs.com/package/formdata-polyfill), but then ran into other issues.

.github/workflows/test.yml
```diff
@@ -13,8 +13,8 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 18.0.0
-          - 19
+          - 20
+          - 21
```
travi, line 17: since 20 will always use the latest version within that range, i always also include a matrix item for the lower bound of the range as well. i'd recommend including 20.8.1 here so that we avoid unintentionally breaking a version that is claimed to be supported

travi (approved):
would love to see the mentioned version added to the matrix, but otherwise, this looks great!

.github/workflows/test.yml
fgreinacher: Good point! Adapted in https://github.com/semantic-release/gitlab/pull/647/commits/52c26906d7bb1772664a0e6508c3e1205868426f

fgreinacher:
Thanks for the feedback @travi, I'll go ahead merge this now 🚀

Issue #636, opened 2023-11-22
Possibility to send succesMessage only for Issue


Hakihiro:
Hi, first, thanks for your work on this project.

Second, I'm guessing if there is any possibility to disable success message but only for issue or MR ?

Because right now it's quite redundant to have the message in the issue and MR. And we can only disable success message for both, but not only for one of them.

If not, is this something you will be open to have in the project.

Sorry if this message is silly, I could not find any information about it.

Hakihiro:
Hi, thanks for your response.

Your proposal seems to be (in my case), what I need.

I don't see this as something complicated either for the utilization or the implementation, are you open to PR or should we wait the approval of the team about it?

JonasSchubert:
Hey @Hakihiro , personally, I would wait for team response before starting to work on that.
There have been some discussions related to similar topics and maybe we can bundle it somehow.

But doesn't mean you shouldn't start working on that or anybody else 😉

fgreinacher:
Thanks for reaching out @Hakihiro!

I think we should first explore the suggestion by @olexs at https://github.com/semantic-release/gitlab/issues/480#issuecomment-1446450041 as this would be most flexible while keeping the configuration surface small.

Feel free to send a PR for this so that we can see how that would feel.

Hakihiro:
Hi, sorry for the delay.

I made it quickly (really inspired by @JonasSchubert suggestion, thanks btw) just to validate the global design. If it's validate by the team, I can add tests and take code feedbacks to improve the modifications, and the documentation ofc.

I think it could be a first step, and it could be overridden later by an information extracted from the comment template.
This way is really the more natural for the user to have an explicit way to configure it, and it can really be easier to understand from a documentation.

I'm open to discuss it.

Thanks again for your time

fgreinacher:
@Hakihiro The new option `successCommentCondition` should do the job 😀

Hakihiro:
Thanks to all contributors for this one, sorry to not following the work, I wasn't able to do it before.

Issue #480, opened 2023-01-11
Allow skipping MR/issue comments based on commit types


nejch:
Not sure if this would be consistent with the GitHub plugin, but it would be cool to be able to filter `getRelatedMergeRequests` based on commit messages in those MRs.

For example, only comment on them and their related issues if their commits contain a release-relevant commit type (feat/fix/breaking change etc), to reduce noise once a release is done, especially for `chore` MRs that bump dependencies and similar.

Something like `"commentOnCommitType": ["feat", "fix", "breaking"]` or so :)

fgreinacher:
I like the idea! @travi @gr2m WDYT?

travi:
we have had requests for similar with the github plugin, such as https://github.com/semantic-release/github/issues/359. it would be ideal to be consistent with github and gitlab if we were to add this functionality. 

since we are in the middle of the esm transition and starting to build a bit more momentum, i would like to get past https://github.com/semantic-release/semantic-release/issues/2133 before giving serious consideration to bringing it into the github plugin. for me to give it full consideration, i think after that happens would be the best time bring it back up for focused consideration again.

nejch:
@travi sure, no rush. Thanks for considering this!

The GitHub plugin issue seems slightly different in that it's more about duplicate comments in non-linear/multi-branch release workflows whereas this would be more about filtering based on type from the get-go. But I understand you'd want to have some kind of consistent approach to adding conditions for creating comments.

olexs:
We have a similar use case in a project, I've been looking at related issues and this is probably the closest short of opening a completely new discussions.

What we would like is to be able to enable or disable `successComment` based on the branch configuration, specifically we'd like to disable the function for pre-releases and keep it enabled for "proper" releases.

The Lodash template function used for `successComment` already can access a lot of necessary metadata, as described here: https://github.com/semantic-release/gitlab#successComment (including the branch, and targeted MR/issue, which would help the OP use case). If there was a way return a value from the template function that basically says "skip posting the comment", this would solve both our cases. This would need to be implemented around https://github.com/semantic-release/gitlab/blob/master/lib/success.js#L32 and https://github.com/semantic-release/gitlab/blob/master/lib/success.js#L48.

fgreinacher:
Sorry for getting back to this that late.

[...]

PR #678, opened 2024-02-13
Feat(Comment): Add new parameters commentSuccessTo

Hakihiro:
Choose between MR or Issue message to be posted. Both is also allowed

Close: https://github.com/semantic-release/gitlab/issues/636

### DIFF
--- a/lib/resolve-config.js
+++ b/lib/resolve-config.js
@@ -45,6 +45,7 @@ export default (
     assets: assets ? castArray(assets) : assets,
     milestones: milestones ? castArray(milestones) : milestones,
     successComment,
+    commentSuccessTo: isNil(commentSuccessTo) ? "merge-request,issue" : commentSuccessTo,
     proxy: getProxyConfiguration(defaultedGitlabUrl, HTTP_PROXY, HTTPS_PROXY, NO_PROXY),
     failTitle: isNil(failTitle) ? "The automated release is failing 🚨" : failTitle,
     failComment,
--- a/lib/success.js
+++ b/lib/success.js
@@ -15,7 +15,7 @@ export default async (pluginConfig, context) => {
     commits,
     releases,
   } = context;
-  const { gitlabToken, gitlabUrl, gitlabApiUrl, successComment, proxy } = resolveConfig(pluginConfig, context);
+  const { gitlabToken, gitlabUrl, gitlabApiUrl, successComment, proxy, commentSuccessTo } = resolveConfig(pluginConfig, context);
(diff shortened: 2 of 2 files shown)


fgreinacher:
Thanks for the PR @Hakihiro, greatly appreciated!

As mentioned in https://github.com/semantic-release/gitlab/issues/480#issuecomment-1446450041 I would prefer to solve this without a new configuration option.

Otherwise we might end with many similar options like `commentSuccessForBranches`, `commentSuccessForStableVersions` and so on :)

JonasSchubert:
> Thanks for the PR @Hakihiro, greatly appreciated!
> 
> As mentioned in [#480 (comment)](https://github.com/semantic-release/gitlab/issues/480#issuecomment-1446450041) I would prefer to solve this without a new configuration option.
> 
> Otherwise we might end with many similar options like `commentSuccessForBranches`, `commentSuccessForStableVersions` and in the worst case we'd need combinations of them 😬

Hi @Hakihiro and thanks a lot for your contribution. I agree with @fgreinacher the risk of configuration overhead is too big and trying a templating attempt might be more sustainable.

Hakihiro:
I understand your point of view. I'm not a big fan to put an information inside a template message to determine on which target(s) it should be posted. But I understand your fear about the unsustainable configuration becoming quickly unmainteable.

I will try to extract this information from the successMessage with lodash template, base on the following syntax:
`[to issues]` - `[to merge-request]` - `[to issues,merge-request]`

Is this something more suitable for you?

Thanks for your feedback

fgreinacher:
Thanks a ton!

> I will try to extract this information from the successMessage with lodash template, base on the following syntax:
> [to issues] - [to merge-request] - [to issues,merge-request]

That's not even needed. The idea would be to not create the note if the rendered template returns a falsey value:

https://github.com/semantic-release/gitlab/blob/master/lib/success.js#L31-L38

Users could then define a template like this to indicate that they only want comments on issues (not sure about the exact syntax, needs some experimentation with the [lodash template engine](https://docs-lodash.com/v4/template/):

```
<% if (issue) { %> regular comment <% } %>
```

JonasSchubert:
Hi @Hakihiro and sorry already as I cross you a bit with another pull request. There was no activity here for some time now, so I created a new [PR](https://github.com/semantic-release/gitlab/pull/730) with [these](https://github.com/semantic-release/gitlab/pull/678#issuecomment-1952644489) suggestions.

fgreinacher:
Closing in favor of #730.

PR #680, opened 2024-02-19
fix(fail): properly parse issue creation response to log in terminal

JonasSchubert:
I realized in our logs, the `id` and `web_url` are not parsed properly. According to the docs the call to the method `.json()` was missing. Local tests showed this fixed the logs and did not affect the normal functionality.

### DIFF
--- a/test/fail.test.js
+++ b/test/fail.test.js
@@ -72,11 +72,16 @@ Your **[semantic-release](https://github.com/semantic-release/semantic-release)*
       labels: "semantic-release",
       title: "The automated release is failing 🚨",
     })
-    .reply(200);
+    .reply(200, { id: 3, web_url: "https://gitlab.com/test_user/test_repo/-/issues/3" });
 
   await fail(pluginConfig, { env, options, branch, errors, logger: t.context.logger });
 
   t.true(gitlab.isDone());
+  t.deepEqual(t.context.log.args[0], [
+    "Created issue #%d: %s.",
+    3,
+    "https://gitlab.com/test_user/test_repo/-/issues/3",
+  ]);
 });
 
 test.serial("Post comments to existing issue", async (t) => {
--- a/lib/fail.js
+++ b/lib/fail.js
@@ -53,10 +53,12 @@ export default async (pluginConfig, context) => {
       debug("create issue: %O", newIssue);
 
       /* eslint camelcase: off */
-      const { id, web_url } = await got.post(issuesEndpoint, {
-        ...apiOptions,
-        json: newIssue,
-      });
+      const { id, web_url } = await got
+        .post(issuesEndpoint, {
+          ...apiOptions,
+          json: newIssue,
+        })
+        .json();
       logger.log("Created issue #%d: %s.", id, web_url);
     }
   }


fgreinacher:
Thanks @JonasSchubert, the changes looks good. Can you please fix the email/signature in your commit so that GitHub is happy? 🙇 

image

JonasSchubert:
> Thanks @JonasSchubert, the changes looks good. Can you please fix the email/signature in your commit so that GitHub is happy? 🙇 
> 
> image
> 

🙈 sorry. work email and private open source email. Will fix that later today or tomorrow latest.

JonasSchubert:
@fgreinacher Updated email and gpg. Thanks for realizing that 🙇

test/fail.test.js
```diff
@@ -72,11 +72,14 @@ Your **[semantic-release](https://github.com/semantic-release/semantic-release)*
...
     })
-    .reply(200);
+    .reply(200, { id: 3, web_url: "https://gitlab.com/test_user/test_repo/-/issues/3" });
 
   await fail(pluginConfig, { env, options, branch, errors, logger: t.context.logger });
 
   t.true(gitlab.isDone());
+  t.assert(t.context.logger.log.getCall(0).args[0], "Created issue #%d: %s.");
+  t.true(t.context.logger.log.getCall(0).args[1] === 3);
+  t.assert(t.context.logger.log.getCall(0).args[2], "https://gitlab.com/test_user/test_repo/-/issues/3");
```
fgreinacher, line 82: Can we do this like the [other similar assertions](https://github.com/semantic-release/gitlab/blob/master/test/verify.test.js#L127)?

```suggestion
  t.assert(t.context.log.args[0], ["Created issue #%d: %s.", 3, "https://gitlab.com/test_user/test_repo/-/issues/3"]);
```

test/fail.test.js
```diff
@@ -1,14 +1,14 @@
...
+import { spy, stub } from "sinon";
 import fail from "../lib/fail.js";
 import authenticate from "./helpers/mock-gitlab.js";
 
 /* eslint camelcase: ["error", {properties: "never"}] */
 
 test.beforeEach((t) => {
   // Mock logger
-  t.context.log = stub();
+  t.context.log = spy();
```
fgreinacher, line 11: If the [suggestion below](https://github.com/semantic-release/gitlab/pull/680/files#r1495396245) works we can keep this as-is I think.

fgreinacher (changes requested):
Thanks @JonasSchubert, had a look at the changes again and maybe we can simplify the test a bit :)

test/fail.test.js
JonasSchubert: `assert` requires a string for comparison, at least this is the error I receive: `"The assertion message must be a string"`. I would change it to `deepEqual`:
```js
t.deepEqual(t.context.log.args[0], ["Created issue #%d: %s.", 3, "https://gitlab.com/test_user/test_repo/-/issues/3"]);
```
Similar to [here](https://github.com/semantic-release/gitlab/blob/master/test/verify.test.js#L107).
Is this ok?

test/fail.test.js
JonasSchubert: I will undo my changes here. `stub` works, too.

test/fail.test.js
fgreinacher: That's what I meant, thanks!

fgreinacher (approved):
Perfect, thanks for the fix!

Issue #704, opened 2024-04-26
Module causes failures in node version 22


aldenbe:
Fails when using node v22 with error: 

```
[semantic-release] › ✘  An error occurred while running semantic-release: SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:337:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:166:18)
    at callTranslator (node:internal/modules/esm/loader:416:14)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:422:30)
SyntaxError: Unexpected identifier 'assert'
```

Caused by: 

https://github.com/nodejs/node/commit/25c79f3331

esm: drop support for import assertions (Nicolò Ribaudo) [#52104](https://github.com/nodejs/node/pull/52104)

https://github.com/semantic-release/gitlab/blob/e890a0b00700e00908cb14db5b091bf1cf10b2e5/lib/definitions/errors.js#L2

ameliendeshams:
Same here

PR #705, opened 2024-04-26
Replaced usage of assert keyword with "with" keyword to support node version 22.0.0

aldenbe:
Fixes https://github.com/semantic-release/gitlab/issues/704

### DIFF
--- a/lib/definitions/errors.js
+++ b/lib/definitions/errors.js
@@ -1,5 +1,5 @@
 import {inspect} from 'util';
-import pkg from '../../package.json' assert {type: 'json'};
+import pkg from '../../package.json' with {type: 'json'};
 
 const [homepage] = pkg.homepage.split('#');
 const linkify = (file) => `${homepage}/blob/master/${file}`;


fgreinacher:
Looks we need to bump our minimum supported Node version to 20.10.0 https://github.com/nodejs/node/pull/50682 because the `with` keyword support was only introduced in https://github.com/nodejs/node/pull/50140.

aldenbe:
> Looks we need to bump our minimum supported Node version to 20.10.0 [nodejs/node#50682](https://github.com/nodejs/node/pull/50682) because the `with` keyword support was only introduced in [nodejs/node#50140](https://github.com/nodejs/node/pull/50140).
> 
> --
> 
> Or maybe even simpler we could just use `require` like the GitHub plugin: https://github.com/semantic-release/github/blob/master/lib/definitions/errors.js#L7

For reference a supported version increase would not have been necessary, this change was not breaking. Support for the "with" keyword was introduced in version 18.20.0 https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#18.20.0 https://nodejs.org/en/blog/release/v18.20.0

I made sure to validate that the change was not breaking before submitting my PR. I'm sorry I did not clarify that better.

fgreinacher:
No worries, that Node.js change was introduced in a quite subtle way... Thanks a ton for reaching out so quickly!

PR #708, opened 2024-04-29
fix: get rid of import assertion

fgreinacher:
Neither import assertions nor the with keyword is supported across our supported versions.

Instead of going for a breaking change and raise our Node dependency I suggest we apply the same logic as [in the GitHub plugin](https://github.com/semantic-release/github/blob/master/lib/definitions/errors.js#L7).

Resolves https://github.com/semantic-release/gitlab/issues/704


travi (approved):

travi:
> Neither import assertions nor the with keyword is supported across our supported versions.

Thanks for confirming that. Absolutely agree given that info :+1:

Issue #529, opened 2023-04-13
An error occurred while running semantic-release: Error: Command failed with exit code 128:


damdafayton:
### Current behavior

I am getting this error too often.

Sometimes if I delete the tags manually it makes new publish. On my last try 5 consecutive trials have failed. Its very annoying. Almost 50% of my release attempts fail by this error and Its not even clear why it fails.

### Expected behavior

make new release

### `semantic-release` version

^17.3.0

### CI environment

gitlab

### Plugins used

    "@semantic-release/git": "^9.0.0",
    "@semantic-release/gitlab": "^6.0.5",
    "@semantic-release/npm": "^7.0.8",

### `semantic-release` configuration

{
  "branches": [
    "master"
  ],
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    [
      "@semantic-release/gitlab",
      {
        "assets": [
          {
            "path": "index.js",
            "label": "Module"
          },
          {
            "path": "README.md",
            "label": "Documentation"
          },
          {
            "path": "LICENSE",
            "label": "License"
          }
        ]
      }
    ],
    "@semantic-release/npm",
    [
      "@semantic-release/git",
      {
        "assets": [
          "package.json"
        ],
        "message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
      }
    ]
  ]
}

### CI logs

`An error occurred while running semantic-release: Error: Command failed with exit code 128: git push [https://gitlab-ci-token:[secure]@gitlab***](https://gitlab-ci-token:%5Bsecure%5D@gitlab***) refs/notes/semantic-release
remote: Retry later
fatal: unable to access 'https://gitlab***': The requested URL returned error: 429
`

fgreinacher:
Status code 429 indicates that you're sending too many requests to GitLab. Is that GitLab instance under your control? If so you could you check which rate limits are configured?

Issue #361, opened 2022-04-11
Response code 429 (Too Many Requests) on successComment


vgropp:
We are running into the gitlab API limit on some projects when multiple semantic-releases are running at the same time from share IPs.

```
[8:07:42 AM] [semantic-release] › ℹ  Start step "success" of plugin "@semantic-release/gitlab"
[8:07:44 AM] [semantic-release] [@semantic-release/gitlab] › ✖  An error occurred while posting comments to related issues and merge requests:
HTTPError: Response code 429 (Too Many Requests)
    at Request.<anonymous> (/usr/local/lib/node_modules/@semantic-release/gitlab/node_modules/got/dist/source/as-promise/index.js:117:42)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'ERR_NON_2XX_3XX_RESPONSE',
```

As far as I understand the code https://github.com/semantic-release/gitlab/blob/3051799160efc127585d30a4ef776ccf34185fee/lib/success.js#L72 for very Commit the script is looking for a MR, with those MR the script is searching for an resolved issue. So there will be at least  commits * MR Requests. As we do not use gitlab issues (but like the MR comment), what about a configuration to disable the Issue Comments and hence those requests? This will limit the requests to count of commits. 

Another Option would be rate limiting the requests, but that seems like a lot more work but more sustainable solution.

fgreinacher:
Thanks for reporting this @vgropp!

You can disable the new feature by setting the `successComment` option to `false`. 

Other than that I think it would make sense to add proper handling of such cases to this plugin.

fgreinacher:
@vgropp Have you seen this behavior recently? In https://github.com/semantic-release/gitlab/pull/728 we learned that the underlying HTTP library does actually have a quite powerful retry logic. I'd like to understand why this is not enough in your case. 🙇

liz72703:
I got this error today.  Never happened before.

AggregateError: 
    HTTPError: Response code 429 (Too Many Requests)
        at Request.<anonymous> (file:///builds/fidus-global/pontem-ui/node_modules/got/dist/source/as-promise/index.js:92:42)
        at Request._onResponseBase (file:///builds/fidus-global/pontem-ui/node_modules/got/dist/source/core/index.js:604:22)
        at async Request._onResponse (file:///builds/fidus-global/pontem-ui/node_modules/got/dist/source/core/index.js:646:13)
[...]

PR #728, opened 2024-06-01
fix(requests): got calls will be retried on response code 429 #361

JonasSchubert:
If got receives a response code 429 - too many requests - it will retry up to three times for GET and POST requests.
This uses the retry API https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md
I am open to change the amount of retries as I selected three as a - for me - good first choice (third time's a charm)


test/publish.test.js
```diff
@@ -641,3 +641,28 @@ test.serial("Publish a release with error response", async (t) => {
...
+  const gitlab = authenticate(env)
+    .post(`/projects/${encodedRepoId}/releases`, {
+      tag_name: nextRelease.gitTag,
+      description: nextRelease.notes,
+      assets: {
+        links: [],
+      },
+    })
+    .reply(429, { message: "Too many requests" })
+    .persist();
```
JonasSchubert, line 663: `.persist` is needed to avoid confusing nock errors.

JonasSchubert:
Might also help to fix https://github.com/semantic-release/gitlab/issues/529

lib/definitions/constants.js
```diff
@@ -1,3 +1,10 @@
 export const HOME_URL = 'https://github.com/semantic-release/semantic-release';
 
 export const RELEASE_NAME = 'GitLab release';
+
+export const RETRY_CONFIGURATION = {
+    retry: {
+        limit: 3,
```
JonasSchubert, line 7: Default in `got` is  `2`. I am open for discussion.

lib/definitions/constants.js
```diff
@@ -1,3 +1,10 @@
 export const HOME_URL = 'https://github.com/semantic-release/semantic-release';
 
 export const RELEASE_NAME = 'GitLab release';
+
+export const RETRY_CONFIGURATION = {
+    retry: {
```
JonasSchubert, line 6: We could try to calculate the `retryAfter` property into optimizing the calculation for the next try.

test/verify.test.js
```diff
@@ -567,7 +567,7 @@ test.serial("Throw error if GitLab API return any other errors", async (t) => {
   const owner = "test_user";
   const repo = "test_repo";
   const env = { GITLAB_TOKEN: "gitlab_token" };
-  const gitlab = authenticate(env).get(`/projects/${owner}%2F${repo}`).times(3).reply(500);
+  const gitlab = authenticate(env).get(`/projects/${owner}%2F${repo}`).times(3).reply(500).persist();
```
JonasSchubert, line 570: .persist is needed to avoid confusing nock errors.

lib/definitions/constants.js
```diff
@@ -1,3 +1,10 @@
...
+        limit: 3,
+        statusCodes: [429],
```
fgreinacher, line 8: I think we can go with the default here:

```suggestion
```

test/publish.test.js
fgreinacher: I'm not sure, did you have a look at https://stackoverflow.com/a/71405383/31985? Maybe we just need `.times(3)` now?

test/verify.test.js
fgreinacher: See [above](https://github.com/semantic-release/gitlab/pull/728/files#r1627183757)

lib/definitions/constants.js
fgreinacher: I think `got` does that already, see https://github.com/sindresorhus/got/blob/cc346b79561b13e3a266b22b7f612fca5a45eb1e/source/core/index.ts#L357

fgreinacher:
Thanks for the PR!

I was not aware that `got` retries failed requests by default, see https://github.com/sindresorhus/got/blob/cc346b79561b13e3a266b22b7f612fca5a45eb1e/documentation/quick-start.md#retries.

Have you specifically noticed issues with the default of 2 retries?

lib/definitions/constants.js
JonasSchubert: Yes indeed. I assumed we are overriding the `retry` options when setting the `apiOptions`. I was wrong. Thanks for properly looking into it. I will close this PR.
What is still confusing though are the reported `429` responses of some users...

lib/definitions/constants.js
JonasSchubert: I will close this PR. Thanks for reviewing!

test/publish.test.js
JonasSchubert: That would work indeed. Thanks for highlighting this.

JonasSchubert (commented):
I will close this PR as `got` is using a default setup for the retries (retrying twice). Wondering now, why some users report `429` if `got` is covering this already.

fgreinacher:
> I will close this PR as `got` is using a default setup for the retries (retrying twice). Wondering now, why some users report `429` if `got` is covering this already.

We still raise the error after retrying after 2 retries, and also https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#maxretryafter might be relevant here.

PR #729, opened 2024-06-02
test(fail): add test cases for failComment or failTitle set to false

JonasSchubert:
This adds three more test cases to fully cover the `fail` functionality. `failTitle` and/or `failComment` set to `false` was not yet covered.


fgreinacher:
Thanks a ton!

fgreinacher (approved):

PR #730, opened 2024-06-02
feat: allow to skip success and fail comments based on provided condition

JonasSchubert:
Could solve #480 and #636. This is based on the idea of @fgreinacher [here](https://github.com/semantic-release/gitlab/pull/678#issuecomment-1952644489).
I added documentation with some highlighting on how to use this functionality and provided some unit tests.

I introduced a new variable `defaultComment` to ease up the filtering for issues and merge requests, but keeping the default comment message.

`semantic-release/github` has at least [one PR](https://github.com/semantic-release/github/pull/360) which shall have the same result, but is a different approach. I like the templating idea @fgreinacher brought up and think this would be a good solution there too - also to keep the plugins similar.

I am aware that this project has an open [PR](https://github.com/semantic-release/gitlab/pull/678). As there was no update since February I took the liberty to create this one, hope this is ok for you @hakihiro.


fgreinacher:
I like it a lot, thanks for picking it up @JonasSchubert!

It's a very flexible approach that does not add much complexity to the code base. 

@travi Any objections on going this way? Would probably be quite easy to handle it the same way in the GitHub plugin.

fgreinacher (approved):

gr2m (approved):
I think the pull request and the feature is great, thank you!

Just one question: it seems that the new `defaultComment` is meant to do two things

1. decide whether to comment at all based on custom logic
2. customize the comment itself? 

In your examples you always use the `<%= defaultComment %>` placeholder, wrapped in an if condition. Wouldn't it make more sense to introduce an option like `commentCondition` in which we can just put in the condition string, using the lodash template syntax?

fgreinacher:
Thanks for the review @gr2m!

It is already possible to disable commenting by statically setting the `successComment` option to `false`. Therefore I think it's more consistent to not handle the use case of dynamic disabling differently.

gr2m:
Sorry if I miss something obvious, but why not do an option like `condition` that can be set to something like `"issue.labels?.includes('semantic-release-relevant')"` instead?

fgreinacher:
We could certainly also do that! My only small concern is that we then have two ways two disable the success comment:

1. set the existing `successComment` to `false`.
2. set the new `commentCondition` option to an expression that evaluates to `false`.

But it's probably not a big issue if we manage to describe it properly :)

gr2m:
independent of what we have today, what would the ideal arguments look like? We shouldn't be afraid of breaking things, it's better than ending up with arguments that are unnecessary complicated due to legacy reasons.

Maybe we could deprecate the possibility of setting `successComment` to `false`, so that it keeps working but we start logging a deprecation message. And we introduce a new argument like `commentCondition` which would be the single place to decide whether a comment done or not?

fgreinacher:
Yes, that sounds like a good plan!

I'm not entirely sure whether `lodash` templating supports evaluating a boolean expression. That would need some investigation.

JonasSchubert:
> We could certainly also do that! My only small concern is that we then have two ways two disable the success comment:
> 
> 1. set the existing `successComment` option to `false`.
> 2. set the new `commentCondition` option to an expression that evaluates to `false`.
> 
> But it's probably not a big issue if we manage to describe it properly :)
> 
> @JonasSchubert WDYT about this idea?

Sounds like a good idea to separate the responsibilities of these configuration options. 👍 We had at least three attempts to filter success comments now (https://github.com/semantic-release/github/pull/360, https://github.com/semantic-release/gitlab/pull/678 and this PR) and this suggestion sounds reasonable.

Should this then also be changed for the `fail` step? It has two flags (`failComment` and `failTitle`) which can either be `false` to disable the `fail` step. Shall this also be handled with a condition or do we keep that logic? I would prefer if we can align the handling of `success` and `fail`.

A proper description is definitely needed, but from experience I know description and documentation might not be read by everybody. 😏 This is why I think we have to make it as understandable as possible without the need for too much documentation.

fgreinacher:
Good point, I agree we should align success and fail comments.

What about `successCommentCondition` and `failCommentCondition`?

gr2m:
> Good point, I agree we should align success and fail comments.
> 
> What about `successCommentCondition` and `failCommentCondition`?

Alternatively we can just make a single `commentCondition`, but expose context to that conditions to differentiate between success and failer?

fgreinacher:
> > Good point, I agree we should align success and fail comments.
> > What about `successCommentCondition` and `failCommentCondition`?
> 
> Alternatively we can just make a single `commentCondition`, but expose context to that conditions to differentiate between success and failer?

I also thought about that, but in the end the reasons for skipping a success comment are quite different from the reasons for skipping a fail comment. The former is usually done to reduce noise, e.g. not spamming irrelevant MRs. The later should IMO only be done in exceptional situations, e.g. when the issue tracker is disabled. I therefore believe mixing both would increase complexity unnecessarily.

travi:
> Would probably be quite easy to handle it the same way in the GitHub plugin (see also [semantic-release/github#359 (comment)](https://github.com/semantic-release/github/issues/359#issuecomment-2143877794))

sorry for my slow responses here. the last few weeks have been especially crazy for me. i appreciate the callout to this detail. i think this is the biggest consideration for me. like @gr2m, i'm open to breakages to get to an api we are comfortable with and think solves the problem well, but would like to make sure we take both github and gitlab into consideration with that design. i still hope that we can eventually extract some of the common behaviors to a more "git-host core" like package to reduce some of the duplication, but that is something that likely is a bit further into the future still.

fgreinacher:
Thanks for all your comments!

Let me summarize:

1. we introduce
   a. _either_ a new option `commentCondition` that can be used to disable both success and fail comments.
   b. _or_ (my favorite) two new options `successCommentCondition` and `failCommentCondition` that can be used to disable success and fail comments respectively
2. we log a warning when people are setting `successComment`. `failComment` or `failTitle` to `false`, announcing this to be removed in future major versions
3. in one of next major releases we can then remove the deprecated functionality

@JonasSchubert Now that the scope has grown a bit are you still willing to work on it?

JonasSchubert:
> Thanks for all your comments!
> 
> Let me summarize:
> 
> 1. we introduce
>    a. _either_ a new option `commentCondition` that can be used to disable both success and fail comments.
>    b. _or_, my favorite, two new options `successCommentCondition` and `failCommentCondition` that can be used to disable success and fail comments respectively
> 2. we log a warning when people are setting `successComment`. `failComment` or `failTitle` to `false`, announcing this to be removed in future major versions
> 3. in one of next major releases we can then remove the deprecated functionality
> 
> @JonasSchubert Now that the scope has grown a bit are you still willing to work on it? 

@fgreinacher I would still be willing to work on it, yes. As soon as we agreed on the way forward I will adjust my PR.

I would then also prefer 1.b. as 1.a. will be very complex to configure for the end user.

fgreinacher:
@JonasSchubert Great, thanks a ton! 

Let's go on with 1b (separate options) then!

gr2m:
summery sounds good, 1b is good with me, thank you 👍🏼

lib/fail.js
```diff
@@ -34,32 +34,39 @@ export default async (pluginConfig, context) => {
     const openFailTitleIssues = await got(openFailTitleIssueEndpoint, { ...apiOptions }).json();
     const existingIssue = openFailTitleIssues.find((openFailTitleIssue) => openFailTitleIssue.title === failTitle);
 
-    if (existingIssue) {
-      debug("comment on issue: %O", existingIssue);
+    const canCommentOnOrCreateIssue = failCommentCondition
```
JonasSchubert, line 37: With this setup we will always have some additional API calls even if we do not want to create an or comment on an issue. Setting `failTitle` or `failComment` to `false` had the advantage of an early stop. This is - currently - not possible and I have no idea if it is possible with `failCommentCondition`.

lib/success.js
```diff
@@ -15,43 +15,62 @@ export default async (pluginConfig, context) => {
...
-        debug("Posting issue note to %s", issueNotesEndpoint);
-        const body = successComment
-          ? template(successComment)({ ...context, issue, mergeRequest: false })
-          : getSuccessComment(issue, releaseInfos, nextRelease);
-        return got.post(issueNotesEndpoint, {
-          ...apiOptions,
-          ...proxy,
-          json: { body },
-        });
+        const canCommentOnIssue = successCommentCondition
```
JonasSchubert, line 34: With this setup we will always have some additional API calls even if we do not want to comment on an issue. Setting `successComment` to `false` had the advantage of an early stop. This is - currently - not possible and I have no idea if it is possible with `successCommentCondition`.

lib/success.js
```diff
@@ -15,43 +15,62 @@ export default async (pluginConfig, context) => {
...
-        debug("Posting MR note to %s", mergeRequestNotesEndpoint);
-        const body = successComment
-          ? template(successComment)({ ...context, issue: false, mergeRequest })
-          : getSuccessComment({ isMergeRequest: true, ...mergeRequest }, releaseInfos, nextRelease);
-        return got.post(mergeRequestNotesEndpoint, {
-          ...apiOptions,
-          ...proxy,
-          json: { body },
-        });
+        const canCommentOnMergeRequest = successCommentCondition
```
JonasSchubert, line 54: see previous comment

test/fail.test.js
```diff
@@ -243,3 +243,130 @@ test.serial("Does not post comments when failComment is set to false", async (t)
...
+        id: 2,
+        iid: 2,
+        project_id: 1,
+        web_url: "https://gitlab.com/test_user/test_repo/issues/2",
+        title: "API should implemented authentication",
+      },
+    ])
+    .post(`/projects/${encodedRepoId}/issues`, {
+      id: "test_user%2Ftest_repo",
+      description: `## :rotating_light: The automated release from the \`main\` branch failed. :rotating_light:
```
fgreinacher, line 331: Can we somehow avoid have that lengthy description here? It's not really important for the test case and hurts a bit on the eyes :) No blocker, just if there's an easy way.

lib/fail.js
```diff
@@ -14,16 +14,16 @@ export default async (pluginConfig, context) => {
...
+  const { gitlabToken, gitlabUrl, gitlabApiUrl, failComment, failTitle, failCommentCondition, labels, assignee } =
+    resolveConfig(pluginConfig, context);
   const repoId = getRepoId(context, gitlabUrl, repositoryUrl);
   const encodedRepoId = encodeURIComponent(repoId);
   const apiOptions = { headers: { "PRIVATE-TOKEN": gitlabToken } };
 
   if (failComment === false || failTitle === false) {
     logger.log("Skip issue creation.");
+    logger.error(`Disabling failure reporting should be done using 'failCommentCondition'.
+Using 'false' for 'failComment' or 'failTitle' is deprecated and will be removed in a future major version!`);
```
fgreinacher, line 26:
```suggestion
    logger.error(`Failure reporting should be disabled via 'failCommentCondition'.
Using 'false' for 'failComment' or 'failTitle' is deprecated and will be removed in a future major version.`);
```

lib/success.js
```diff
@@ -15,43 +15,62 @@ export default async (pluginConfig, context) => {
...
+    context
+  );
   const repoId = getRepoId(context, gitlabUrl, repositoryUrl);
   const encodedRepoId = encodeURIComponent(repoId);
   const apiOptions = { headers: { "PRIVATE-TOKEN": gitlabToken } };
 
   if (successComment === false) {
     logger.log("Skip commenting on issues and pull requests.");
+    logger.error(`Disabling commenting on issues and pull requests should be done using 'successCommentCondition'.
+Using 'false' for 'successComment' is deprecated and will be removed in a future major version!`);
```
fgreinacher, line 29:
```suggestion
    logger.error(`Commenting on issues and pull requests should be disabled via 'successCommentCondition'.
Using 'false' for 'successComment' is deprecated and will be removed in a future major version.`);
```

fgreinacher (commented):
Thanks a ton for the rework @JonasSchubert and thanks for your patience.

I like it a lot, just some smaller thoughts.

lib/fail.js
JonasSchubert: Do I understand you correct, that your suggestion is, to call 
```ts
const canCommentOnOrCreateIssue = failCommentCondition
  ? template(failCommentCondition)({ ...context })
  : true;
```
also before
```ts
  if (failComment === false || failTitle === false) {
```
and add
```ts
  if (failComment === false || failTitle === false || canCommentOnOrCreateIssue  === false) {
```
or change it to
```ts
  const isFailureReportDisabled = failComment === false ||
    failTitle === false ||
    canCommentOnOrCreateIssue  === false;
  if (isFailureReportDisabled) {
```
?

We would have to call it twice then:
1. first without the potentially existing issue
2. then with the issue if already existing

Same would apply to success handling. Or do I miss something?

I am totally fine with calling it twice to cover every case and also reduce the API calls, I just want to have it discussed as I am not sure with the best approach.

fgreinacher: @JonasSchubert Evaluating the template without issue/MR might be problematic depending on the logic that folks bake in there.

I would really only do the optimization for the simplest case where `{success,fail}CommentCondition` is statically set to `false`.

README.md
```diff
@@ -146,6 +148,29 @@ The message for the issue comments is generated with [Lodash template](https://l
...
+| `lastRelease`  | `Object` with `version`, `channel`, `gitTag` and `gitHead` of the last release.                                                      |
+| `nextRelease`  | `Object` with `version`, `channel`, `gitTag`, `gitHead` and `notes` of the release being done.                                       |
+| `commits`      | `Array` of commit `Object`s with `hash`, `subject`, `body` `message` and `author`.                                                   |
+| `releases`     | `Array` with a release `Object`s for each release published, with optional release data such as `name` and `url`.                    |
+| `issue`        | A [GitLab API Issue object](https://docs.gitlab.com/ee/api/issues.html#single-issue) the comment will be posted to.                  |
+| `mergeRequest` | A [GitLab API Merge Request object](https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr) the comment will be posted to. |
+
+##### successCommentCondition example
+
+- do no create any comments at all: set to `false` or templating: `"<% return false; %>"`
```
gr2m, line 167:
```suggestion
- do not create any comments at all: set to `false` or templating: `"<% return false; %>"`
```

README.md
```diff
@@ -164,6 +189,27 @@ The `failComment` `This release from branch ${branch.name} had failed due to the
...
+| `branch`      | `Object` with `name`, `type`, `channel`, `range` and `prerelease` properties of the branch from which the release is done.                                   |
+| `lastRelease` | `Object` with `version`, `channel`, `gitTag` and `gitHead` of the last release.                                                                              |
+| `nextRelease` | `Object` with `version`, `channel`, `gitTag`, `gitHead` and `notes` of the release being done.                                                               |
+| `commits`     | `Array` of commit `Object`s with `hash`, `subject`, `body` `message` and `author`.                                                                           |
+| `releases`    | `Array` with a release `Object`s for each release published, with optional release data such as `name` and `url`.                                            |
+| `issue`       | A [GitLab API Issue object](https://docs.gitlab.com/ee/api/issues.html#single-issue) the comment will be posted to - only available if an open issue exists. |
+
+##### failCommentCondition example
+
+- do no create any comments at all: set to `false` or templating: `"<% return false; %>"`
```
gr2m, line 207:
```suggestion
- do not create any comments at all: set to `false` or templating: `"<% return false; %>"`
```

gr2m (approved):
Just found two typos, other than that this is good to go!

And just making sure, we do not introduce breaking changes, only log a deprecation warning?

Thank you all again for putting so much thought and work into making this feature great

fgreinacher (approved):
Really great work @JonasSchubert, I love it!

fgreinacher:
Hey @JonasSchubert, we noticed that you have been super active here for quite some time now and brought in multiple high-quality and impactful contributions. 

We'd therefore like to invite you as an additional maintainer for the GitLab plugin.

Would that be interesting for you?

/cc @travi

travi:
thanks for the help, @JonasSchubert! i've invited you to our gitlab team

JonasSchubert:
Thanks a lot @. Appreciate all your work and I am happy to be psrt of the team!

fgreinacher:
Welcome @JonasSchubert to the team, awesome to have more people here 🥳

Issue #753, opened 2024-07-23
Files uploaded using assets with default target project_upload are not downloadable using link created in release


aldenbe:
The link generated when publishing a release is in the format 

`https://gitlab.com/-/<project path>/uploads/<upload id>/<file>` however files are available at `https://gitlab.com/-/project/<project id>/uploads/<upload id>/<file>`

This causes links in releases to return 404

JonasSchubert:
Hi @aldenbe and thanks for you issue and PR.
Can you provide a example repository or the settings this repository uses. Especially whether you configured a [repository](https://semantic-release.gitbook.io/semantic-release/usage/configuration#repositoryurl) in the `package.json`?
And also which Gitlab version and semantic release version (and plugin versions) you are using?

aldenbe:
> Hi @aldenbe and thanks for you issue and PR. Can you provide a example repository or the settings this repository uses. Especially whether you configured a [repository](https://semantic-release.gitbook.io/semantic-release/usage/configuration#repositoryurl) in the `package.json`? And also which Gitlab version and semantic release version (and plugin versions) you are using?

The entirety of my current gitlab interactions are based around a private group with a lot of my environment tailored to interacting with that group specifically. It may not be as frustrating for me to demonstrate this publicly as I would guess, but it would take me a decent amount of time to be willing to find out.

Otherwise https://github.com/semantic-release/gitlab/issues/750#issuecomment-2250436652 seems to suggest others are experiencing the same result, and while https://docs.gitlab.com/ee/api/projects.html#markdown-uploads makes claims that in my testing appear to not be true, my solution does line up as valid given their documented api response.

aldenbe:
Also, it might be worth updating the readme to mention that the default target (on gitlabs end too) of direct uploads is just all around terrible, and either the package registry or external file hosting should be used if possible.

aodendaal:
I have a [public demo repo](https://gitlab.com/andre3254704/test-demo) you are welcome to use which I have for testing semantic release and which demonstrates the recent bug with the Changelog giving 404.

fgreinacher:
> I have a [public demo repo](https://gitlab.com/andre3254704/test-demo) you are welcome to use which I have for testing semantic release and which demonstrates the recent bug with the Changelog giving 404.

Thank you @aodendaal :thum

PR #754, opened 2024-07-23
fix(assets): changed generation of gitlab release asset url

aldenbe:
fix(assets): changed generation of gitlab release asset url when providing file assets with target project_upload to point to url for the uploaded file

#753


lib/publish.js
```diff
@@ -149,7 +149,8 @@ export default async (pluginConfig, context) => {
               throw error;
             }
 
-            const { url, alt } = response;
+            const { alt } = response;
+            const url = urlJoin(gitlabUrl, response["full_path"]);
```
JonasSchubert, line 153: This breaks the unit tests. Can you locally run `npm run test`.

```suggestion
            const url = urlJoin(gitlabUrl, response.full_path);
```

In `ln. 177` we are already building the url for the assets using `urlJoin(gitlabUrl, repoId, url)`.

JonasSchubert (changes requested):
Let's also discuss a bit in the [issue](https://github.com/semantic-release/gitlab/issues/753#issuecomment-2248692476) ;)

test/publish.test.js
```diff
@@ -56,7 +56,7 @@ test.serial("Publish a release with templated path", async (t) => {
   const encodedGitTag = encodeURIComponent(nextRelease.gitTag);
   const generic = { path: "${env.FIXTURE}.txt", filepath: "/upload.txt" };
   const assets = [generic];
-  const uploaded = { url: "/uploads/upload.txt", alt: "upload.txt" };
+  const uploaded = getUploadResponse(generic.filepath, encodedRepoId);
```
fgreinacher, line 59: I would prefer to keep the test cases mostly self-contained and hardcode the upload responses like we do for most other test data.

fgreinacher (changes requested):
Thanks for your contribution @aldenbe, greatly appreciated!

The fix itself looks perfect, but I'd prefer to keep the existing test structure

lib/publish.js
aldenbe: see https://github.com/semantic-release/gitlab/pull/754/commits/1c48f54e47ffd47fba3fa172e77c75f7fc4f738b#diff-c8bd325917d181608f40adecdcec3c76acde134cc15a9e4a0d18636698e4fe55R152-R153

matching the previously used style

test/publish.test.js
aldenbe: It is my opinion that using a helper function to mimic the gitlab api endpoint responses offers the following benefits over hardcoded values.

- Increased readability. It is more clear what determines the url values passed to the simulated release endpoint which are programatically generated by gitlab and not semantic-release it also removes the implication of some variables being treated as a gitlab response which are not actually contained in response data from the gitlab markdown upload endpoint. see [parameters expected in release request taken from variable uploaded not supplied by gitlab markdown upload endpoint](https://github.com/semantic-release/gitlab/pull/754/commits/35c76790fbe11d88f8c968c180a062964e562e21#diff-e5529dd6d95cc6412d41048a450c6d2b3647beaba2e5489a02094ab5b4441d4fL334-R336) and [uploaded implied to be a response from the gitlab markdown upload endpoint](https://github.com/semantic-release/gitlab/pull/754/commits/35c76790fbe11d88f8c968c180a062964e562e21#diff-e5529dd6d95cc6412d41048a450c6d2b3647beaba2e5489a02094ab5b4441d4fR342-R344)
- Increased maintainability. Should gitlab change their endpoint (again) updates need only to be made to [the mocked gitlab upload endpoint](https://github.com/semantic-release/gitlab/pull/754/commits/d4e057647f31bfb929f28a8c283ed96394aa9836#diff-9613b6e899c4d22ecbc1d582369aa824b4e77744ec607b5addacb9e65fb55f24R46-R58) instead of each test case being updated individually.
- Increased reliability. Although no test case currently exists, breakages potentially caused by the gitlab upload behavior of [only using the file name and ignoring the file path](https://github.com/semantic-release/gitlab/pull/754/commits/d4e057647f31bfb929f28a8c283ed96394aa9836#diff-9613b6e899c4d22ecbc1d582369aa824b4e77744ec607b5addacb9e65fb55f24R46-R48) would be more easily detected by mimicking the gitlab endpoint response than relying on hardcoded values and expecting developers to properly account for this, especially given that as far as I can tell this behavior is not actually documented anywhere (including in gitlab's documentation).

aldenbe:
> Thanks for your contribution @aldenbe, greatly appreciated!
> 
> The fix itself looks perfect, but I'd prefer to keep the existing test structure.
> 
> Also we have some lint issues, see https://github.com/semantic-release/gitlab/actions/runs/10120894480/job/28034084391?pr=754. You should be able to fix them via `npm run lint:fix`.

https://github.com/semantic-release/gitlab/compare/1d4d17336b77aaa8ecdc58174cfcacda12e673be..35c76790fbe11d88f8c968c180a062964e562e21

test/publish.test.js
aldenbe: [gitlab response handling in publish.js](https://github.com/semantic-release/gitlab/blob/d53f94c391792f81bc27528a428ca561cd8fb905/lib/publish.js#L146) would also contain the unexpected variables in point 1, which makes that point both a readability and reliability concern.

fgreinacher: Thanks for providing you reasoning @aldenbe, it's very clear!

I do agree with you in general, but seeing that we have many other similar places where we "hardcode" responses, e.g. https://github.com/semantic-release/gitlab/blob/4554f49a34e064fd26c8bd968f83841ee81ee914/test/publish.test.js#L181-L182 I would prefer to keep that style _in this PR_. I would be very open to accept a PR that refactors all the responses towards the style you suggest here. Would you be willing to give that a try?

@JonasSchubert WDYT?

aldenbe: Simplified PR created explicitly addressing this issue created https://github.com/semantic-release/gitlab/pull/755

Unfortunately I have spent too much time on this already. Feel free to take anything desired from https://github.com/aldenbe/semantic-release-gitlab/tree/gitlab-mock-upload-endpoint and use it as you please.

JonasSchubert: Thanks again @aldenbe for your work and fix.
I agree with @fgreinacher to keep the context of PRs to one topic to avoid too many changes in one. This usually helps to review a PR. I would be happy to see you create a second PR with your changes for the tests, @aldenbe. This can indeed improve readability. I understand if you have no time for this now, but if you would like to tackle that in the future it would be appreciated.

PR #799, opened 2024-12-02
fix: updated configuration example

gaidzla:
Add assets label for url otherwise release is not created because GitLab API expect a assets:links:name.

https://docs.gitlab.com/ee/api/releases/#create-a-release


fgreinacher (approved):
Thank you!

PR #800, opened 2024-12-03
fix: increase retry limit and make it configurable

### LINKED ISSUE #361: Response code 429 (Too Many Requests) on successComment

fgreinacher:
Resolves #361


fgreinacher:
@JonasSchubert Could you have a look here? 🙇

JonasSchubert (approved):
LGTM.

PR #813, opened 2025-01-24
feat(publish): Add support for future or historical releases with new releasedAt parameter

thossullivan:
Howdy! Would love feedback on this PR. Happy to adjust based on feedback and definitely let me know if I missed anything!

## Support for Historical // Future Releases

This PR adds support for scheduled releases in GitLab using the `releasedAt` parameter. This enhancement allows users to schedule releases for future dates or create historical releases by leveraging GitLab's release API `released_at` parameter.

## Problem
Currently, the semantic-release GitLab plugin creates releases immediately upon execution. There's no built-in way to schedule a release for a future date or specify a historical release date, which limits release planning and historical record keeping.

## Solution
This PR implements support for GitLab's `released_at` parameter, allowing users to:
- Schedule releases for future dates
- Create releases with historical dates
- Maintain the default immediate release behavior when no date is specified

The implementation adds a new optional configuration parameter `releasedAt` that accepts an ISO 8601 formatted date string.

## Example Usage
```json
{
  "plugins": [
      "releasedAt": "2025-02-01T12:00:00Z"  // Schedule release for February 1st, 2025
    }]
  ]
}
```
## Implementation Details

1. Added releasedAt parameter to the plugin configuration
2. Modified the release payload to include released_at when specified

## Added test coverage including:

1. Scheduled future releases
2. Historical releases
3. Releases with assets and releasedAt
4. Null handling for releasedAt

## Testing

- Added unit tests covering various release scenarios
- Manually tested against GitLab instance to verify proper scheduling behavior
- Verified backward compatibility with existing release functionality

Screenshot 2025-01-24 at 3 39 58 PM

## Linting
- This did pass the linting locally
- prettier --check "*.{js,json,md}" ".github/**/*.{md,yml}" "{bin,lib,test}/*.js"

## Documentation

- Updated README.md with new parameter documentation
- Added usage examples for scheduled releases
- Included best practices and formatting requirements

### Future Release Example
Screenshot 2025-01-24 at 3 31 45 PM

### Historical Release Example
Screenshot 2025-01-24 at 3 35 30 PM

### Example with explicit `null` value
Screenshot 2025-01-24 at 3 36 59 PM


travi:
while interesting, i dont see this being a fit for use with semantic-release. semantic-release has always encouraged a fully automated, continuous-deployment style workflow, which doesnt align well to scheduled releases

thossullivan:
Thank you @travi for taking the time to review this PR and provide thoughtful feedback! I completely respect the commitment to fully automated, continuous deployment workflows.

I'd like to share the specific use case that inspired this feature: When helping organizations transition from traditional release management toward CD, there is sometimes a need to bridge the gap.

For those teams transitioning from gitflow --> CD, they use release branches to stage and validate changes before production deployment, while working toward the goal of true CD and trunk-based development.

In these transitional scenarios, the ability to create a scheduled release would allow teams to:
1. Generate comprehensive release notes early using semantic-release's thorough automation
2. Update those notes automatically if additional commits are added to the release branch
3. Finally update the release timestamp to the actual deployment time when merging to main

However, I understand that supporting such transitional workflows may not align with semantic-release's core philosophy and goals. I appreciate you helping maintain the project's clear focus on CD best practices.

Happy to close this PR. Thank you again for the review and for maintaining such a valuable tool for the community!

travi:
the other trouble i see with this is that this is limited to the gitlab api, so if you are publishing an npm package, for example, and publishing the release notes to gitlab, the npm package would publish at the point that semantic-release executes, but the release notes would be delayed from posting until the scheduled time? that sounds like it would only cause confusion and not actually solve what you describe.

i think it would be a far better option to consider scheduling the merge of the branch into the release branch. i know of some options for doing this with github, but i'm personally less familiar with gitlab, so i dont have a specific solution to offer. either way, i think any sort of scheduling along those lines is outside the scope of responsibility of semantic-release. once the merge happens, semantic-release will happily do its job and coordinate the various parts of the release at that point

thossullivan:
Thanks for this thoughtful explanation! It really helps me understand the bigger picture concern about keeping package releases and documentation in sync.

I was envisioning using the scheduled release notes more as a preview feature - when commits hit the release branch, we'd generate and schedule release notes to show what's coming to our users, but the actual package publishing would still happen on merge to main with updated "real-time" release notes.

But you make a great point that this kind of preview functionality probably fits better in GitLab's branch management and CI/CD pipeline features rather than in semantic-release itself. Managing timing at the branch level is a much cleaner approach.

This has been super helpful in understanding semantic-release's architecture better!

Appreciate the time, @travi!

PR #842, opened 2025-04-08
feat: add option to publish to CI catalog

fgreinacher:


fgreinacher:
@JonasSchubert WDYT about doing it like this as a first step?

README.md
```diff
@@ -211,6 +212,14 @@ The fail comment condition is generated with [Lodash template](https://lodash.co
 
 > check the [GitLab API Issue object](https://docs.gitlab.com/ee/api/issues.html#single-issue) for properties which can be used for the filter
 
+#### publishToCatalog
+
+**Note**: This is an EXPERIMENTAL option that might change in the future.
```
travi, line 217: if considering this experimental, is it worth publishing as an alpha or beta release to get feedback from interested folks before releasing in a stable version?

fgreinacher: Yep, good point. I have retargeted this one to `next`.

lib/publish.js
```diff
@@ -201,5 +193,15 @@ export default async (pluginConfig, context) => {
 
   const releaseUrl = urlJoin(gitlabUrl, projectPath, `/-/releases/${encodedGitTag}`);
 
+  if (publishToCatalog) {
+    try {
+      await execa("glab", ["repo", "publish", "catalog", gitTag], { cwd, timeout: 30 * 1000 });
```
JonasSchubert, line 198: - This also introduces `execa` as a new dependency. What about just using `exec` from `node:child_process`? (`execa` is also not yet listed in `package.json` or here as import)
- Also `glab` respects `GITLAB_TOKEN` according to their [docs](https://docs.gitlab.com/editor_extensions/gitlab_cli/#authenticate-with-gitlab). But the plugin allows `GL_TOKEN` or `GITLAB_TOKEN`.  We should either add a check here and/or highlight this in the README.

fgreinacher: Thanks for the feedback @JonasSchubert!

Good point, I just saw `execa` used in other `semenatic-release` repos. I'll try with the standard lib instead.

Also good hint with the token, I think we need to pass on our variable to the child process.

I'll be out for the next two weeks, so don't expect any activity here :)

README.md
fgreinacher: Based on the feedback from https://github.com/semantic-release/gitlab/pull/843 I have retargeted this to a fresh `beta` branch.

lib/publish.js
fgreinacher: > Good point, I just saw execa used in other semenatic-release repos. I'll try with the standard lib instead.

Just had another look. We're using [`execa` nearly everywhere](https://github.com/search?q=org%3Asemantic-release+execa%28&type=code), so let's go for consistency here.

fgreinacher: > Also glab respects GITLAB_TOKEN according to their [docs](https://docs.gitlab.com/editor_extensions/gitlab_cli/#authenticate-with-gitlab). But the plugin allows GL_TOKEN or GITLAB_TOKEN. We should either add a check here and/or highlight this in the README.

I have adapted the `execa` call to pass the token via the `GITLAB_TOKEN` variable.

lib/publish.js
```diff
@@ -201,5 +193,19 @@ export default async (pluginConfig, context) => {
...
+        cwd,
+        timeout: 30 * 1000,
+        env: { GITLAB_TOKEN: gitlabToken },
+      });
+      logger.log("Published tag %s to the CI catalog", gitTag);
+    } catch (error) {
+      logger.error("An error occurred while publishing tag %s to the CI catalog:\n%O", gitTag, error);
+      throw error;
+    }
+  }
```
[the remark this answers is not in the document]
fgreinacher: I agree @JonasSchubert, I think the best we could do is expect the tool to be called with the right arguments and environment variables, but I'm also not sure it's worth adding the complexity.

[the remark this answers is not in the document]
travi: i think it would be valuable to at least stub/mock and make sure we are passing the expected inputs, but exercising it fully would require a more end-to-end setup like we have with docker for core or the npm plugin. this doesnt feel like the trigger to go that far, but worth being aware of as an option we've used elsewhere. we havent gone that far yet with the github plugin either, but it is exercised in the e2e tests for core against a fake api since it is a default plugin.

xrow:
@fgreinacher 

I have a wish to this pull. It is possible to autodetect if this a components project by reading the checkbox via api? Of course the config flag shall be present with true false being autodetected if not set. The default value could be maybe null or auto.

image

lib/publish.js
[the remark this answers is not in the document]
fgreinacher: Thanks for the feedback. I have added some basic stub testing!

fgreinacher:
With the tests in place I think this is now ready for review.

Can you have a look again @JonasSchubert? 🙇‍♂️

fgreinacher:
> @fgreinacher
> 
> I have a wish to this pull. It is possible to autodetect if this a components project by reading the checkbox via api? Of course the config flag shall be present with true false being autodetected if not set. The default value could be maybe null or auto.
> 
> image

Good idea! I have added it to https://github.com/semantic-release/gitlab/issues/851 to track this.

xrow:
@fgreinacher 

I did try to verify the changes to this change, but I was not able to test this successfully. I ended up with error 
`Response code 422 (Release cannot contain more than 30 components)` again.

The errror with debug: https://gitlab.com/xrow-public/ci-tools/-/jobs/10053788956#L517
The config: https://gitlab.com/xrow-public/ci-tools/-/blob/main/.releaserc.json?ref_type=heads
The package.json: https://gitlab.com/xrow-public/ci-tools/-/blob/main/container/semantic-release/package.json?ref_type=heads#L11

It might be something on my end, but I can`t tell what went wrong

JonasSchubert (commented):
LGTM.
Were not able to run and test it like @xrow.

fgreinacher:
> I did try to verify the changes in this pull, but I was not able to test this successfully. I ended up with error
> `Response code 422 (Release cannot contain more than 30 components)` again.

@xrow This is a [general limitation of GitLab CI components](https://docs.gitlab.com/ci/components/#component-project):

> [...], with a maximum of 30 components per project.

fgreinacher:
Thanks all for the feedback! I'm gonna merge this so that we have `beta` release to test with.

PR #843, opened 2025-04-10
Chore/update next

fgreinacher:


travi:
typically, `next` / `alpha` / `beta` are not intended to last beyond the process of getting a release promoted to stable by merging to the default branch. after that, it is best to just delete the next or pre-release branch and create a new one from the default branch when a new one is needed

travi:
also, out of curiosity, is there a reason you are preferring the next workflow for this rather than a pre-release workflow? i'm genuinely curious since i have not found the next workflow to be useful in my normal process, but use pre-releases a fair amount. 

next will publish using the versions that continue from the primary release channel, but not tag as latest. pre-releases will publish to separate channels and use pre-release versioning, so the version numbers dont conflict with the default channel. that way the pre-release can be released and evaluated in parallel with other work that might continue forward in the default channel.

fgreinacher:
> also, out of curiosity, is there a reason you are preferring the next workflow for this rather than a pre-release workflow? i'm genuinely curious since i have not found the next workflow to be useful in my normal process, but use pre-releases a fair amount.
> 
> next will publish using the versions that continue from the primary release channel, but not tag as latest. pre-releases will publish to separate channels and use pre-release versioning, so the version numbers dont conflict with the default channel. that way the pre-release can be released and evaluated in parallel with other work that might continue forward in the default channel.

(Sorry for the late reply, just back from vacation)

That's a good hint, thanks @travi! I was actually not aware of this distinction. So in this case I will create a new `beta` branch and publish from there!

Issue #756, opened 2024-08-02
consider switching from got to ky


travi:
https://www.npmjs.com/package/ky

from the `got` readme:
> You probably want [Ky](https://github.com/sindresorhus/ky) instead, by the same people. It's smaller, works in the browser too, and is more stable since it's built upon [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

i've used ky in a few projects now and it has been solid. would it make sense to switch here?

fgreinacher:
Sure, why not. 

What comes immediately to my mind when switching HTTP libraries is whether the properly support proxy env variables. That is often causing issues. Do you have experience with that?

JonasSchubert:
I am going to test it. I switched a package from got to ky and going to monitor it. Proxy settings can be forwarded to the node fetch method as far as I understood it.
Will come back here, once I have tested more.

movies0:
I am going to test the changes. I have switched a package from got to ky and will monitor the results. As far as I understand, proxy settings can be forwarded to the node-fetch method.

I will update here once I have conducted more tests.

Let me know if you'd like any further tweaks!

JonasSchubert:
> I am going to test it. I switched a package from got to ky and going to monitor it. Proxy settings can be forwarded to the node fetch method as far as I understood it. Will come back here, once I have tested more.

After using my updated package now for a few days, here a short summary:
- theoretically the proxy settings can be forwarded to node-fetch, thanks to [this PR](https://github.com/sindresorhus/ky/pull/536)
- ky will [not receive builtin proxy support](https://github.com/sindresorhus/ky/issues/526#issuecomment-1739053745) though
- I (naively) have tried the [current approach](https://github.com/semantic-release/gitlab/blob/master/lib/resolve-config.js#L135-L168) also in my lib, but it did not work (probably expected)
- I have successfully used [this approach](https://github.com/sindresorhus/ky/issues/526#issuecomment-2390430300) with undici's **experimental** [EnvHttpProxyAgent](https://github.com/nodejs/undici/blob/main/docs/docs/api/EnvHttpProxyAgent.md)

As the [EnvHttpProxyAgent](https://github.com/nodejs/undici/blob/main/docs/docs/api/EnvHttpProxyAgent.md) is still **experimental** I do not think we should use this in this library. Mine is a minor one without widespread usage, but `semantic-release` should be stable and not use an experimental library.

[...]

PR #844, opened 2025-04-11
refactor(api): use ky and undici instead of got #756

JonasSchubert:
Initial draft to move from got to ky (using undici).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Create a [project access token](https://docs.gitlab.com/ee/user/project/settings
 
 #### Proxy configuration
 
-The plugin supports passing requests through a proxy server.
+The plugin supports passing requests through a proxy server using `undici`'s `EnvHttpProxyAgent`.
 
 You can configure a proxy server via the `HTTPS_PROXY` environment variable: `HTTPS_PROXY=http://proxyurl.com:8080`
 
--- a/lib/fail.js
+++ b/lib/fail.js
@@ -1,6 +1,7 @@
+import { EnvHttpProxyAgent } from "undici";
 import { template } from "lodash-es";
 import urlJoin from "url-join";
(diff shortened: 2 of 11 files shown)


JonasSchubert:
- Tests are failing caused by `undici`s minimum node version being set to `20.18.1` in this [PR](https://github.com/nodejs/undici/pull/3880).
- either we also introduce a breaking change by updating our engine to same `20.x.x` or higher or we abandon that for now
- also @ievgennaida made a good point, but instead of mixing `undici` and `ky` we could simply use only `undici` (if we continue this direction)

travi:
> * we also introduce a breaking change by updating our engine to same `20.x.x` or higher

with node v18 going EOL at the end of the month, it is a good time to consider tightening our supported ranges. however, [we normally](https://semantic-release.gitbook.io/semantic-release/support/node-support-policy) change our support to be limited to only the latest LTS line when we do release a breaking chain, so that goes beyond even the change to undici. 

> we could simply use only undici

i assume we would then use it through the native fetch inerface built into node rather than using undici as a dependency at that point, right? in that case, the node version requirement seems to go away

i guess the question comes down to whether we see ky as providing some value beyond native fetch or not. i dont have strong feelings either way as long as the path we choose favors maintainability.

ievgennaida:
`fetch` is introduced in v18 as experimental and became stable in Node.js v21.

JonasSchubert:
> i assume we would then use it through the native fetch inerface built into node rather than using undici as a dependency at that point, right? in that case, the node version requirement seems to go away
> 
> i guess the question comes down to whether we see ky as providing some value beyond native fetch or not. i dont have strong feelings either way as long as the path we choose favors maintainability.

I will continue working on it this week and would try to use node's built-in functionality.

PR #846, opened 2025-04-22
Handle ci/cd job token

AlexandreEXFO:
Handle ci/cd job token to allow the use of the Gitlab CI/CD token having push rights in the repository (incoming feature)

- https://gitlab.com/gitlab-org/gitlab/-/issues/475705#note_2461866773
- https://gitlab.com/gitlab-org/gitlab/-/issues/468320

Related to https://github.com/semantic-release/gitlab/issues/156

### DIFF
--- a/lib/get-token-header.js
+++ b/lib/get-token-header.js
@@ -1,0 +1,8 @@
+export default (token) => {
+    switch(true) {
+        case token.startsWith("glcbt-"):
+          return "JOB-TOKEN"
+        default:
+          return "PRIVATE-TOKEN"
+      }
+};
--- a/lib/verify.js
+++ b/lib/verify.js
@@ -54,17 +54,25 @@ export default async (pluginConfig, context) => {
   }
 
   if (gitlabToken && projectPath) {
+    let tokenHeader;
     let projectAccess;
     let groupAccess;
 
     logger.log("Verify GitLab authentication (%s)", gitlabApiUrl);
+
+    switch(true) {
+      case str.startsWith("glcbt-"):
+        tokenHeader = "JOB-TOKEN"
+      default:
+        tokenHeader = "PRIVATE-TOKEN"
+    }
 
     try {
       ({
         permissions: { project_access: projectAccess, group_access: groupAccess },
       } = await got
         .get(projectApiUrl, {
-          headers: { "PRIVATE-TOKEN": gitlabToken },
+          headers: { tokenHeader: gitlabToken },
           ...proxy,
         })
         .json());
--- a/lib/verify.js
+++ b/lib/verify.js
@@ -6,6 +6,7 @@ import AggregateError from "aggregate-error";
 import resolveConfig from "./resolve-config.js";
 import getProjectContext from "./get-project-context.js";
 import getError from "./get-error.js";
+import getTokenHeader from "./get-token-header.js"
 
 const isNonEmptyString = (value) => isString(value) && value.trim();
 const isStringOrStringArray = (value) =>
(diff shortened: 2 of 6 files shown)


lib/verify.js
```diff
@@ -54,17 +54,25 @@ export default async (pluginConfig, context) => {
...
     let groupAccess;
 
     logger.log("Verify GitLab authentication (%s)", gitlabApiUrl);
 
+    switch(true) {
+      case str.startsWith("glcbt-"):
+        tokenHeader = "JOB-TOKEN"
+      default:
+        tokenHeader = "PRIVATE-TOKEN"
+    }
```
toms-place, line 68:
```suggestion
    switch(gitlabToken) {
      case gitlabToken.startsWith("glcbt-"):
        tokenHeader = "JOB-TOKEN"
      default:
        tokenHeader = "PRIVATE-TOKEN"
    }
```

@AlexandreEXFO - you need to switch the gitlabToken here.

lib/get-token-header.js
```diff
@@ -0,0 +1,8 @@
+export default (token) => {
+    switch(true) {
+        case token.startsWith("glcbt-"):
+          return "JOB-TOKEN"
+        default:
+          return "PRIVATE-TOKEN"
+      }
```
toms-place, line 7:
```suggestion
    if (token.startsWith("glcbt-")) return "JOB-TOKEN";
    return "PRIVATE-TOKEN";
```

I don't really understand the need to use `switch(true)` here?

AlexandreEXFO:
@travi: Would you have time to review this MR? Thanks in advance!

fgreinacher:
Even if the job token can be used to push to the repository it would still fail for the other API calls, right?

lib/get-token-header.js
AlexandreEXFO: @fgreinacher: It's more complicated to support... We would have to add configuration parameters to support this. Currently, there is no regression with the change since if the prefix has been modified, it will continue to fall into the default case with it's the current implementation. It might be necessary to add a note in the documentation that currently CI job tokens with custom prefixes are not supported. I hadn't really planned on making the changes to support custom prefixes.

AlexandreEXFO:
@fgreinacher: It's possible that there will be failures when calling other methods. It will depend on the CI job token's access to call these methods, but since the majority of calls except for push are read-only, I don't anticipate any. I haven't been able to test the modifications yet since the push feature with CI job token is still under development on Gitlab's side.

fgreinacher:
> @fgreinacher: It's possible that there will be failures when calling other methods. It will depend on the CI job token's access to call these methods, but since the majority of calls except for push are read-only, I don't anticipate any. I haven't been able to test the modifications yet since the push feature with CI job token is still under development on Gitlab's side.

IIUC the linked MRs are just about allowing Git push operations with the job token. We still won't be able to use it for the API calls which are needed for this plugin to work.

AlexandreEXFO:
@fgreinacher : It's already possible to call multiple APIs with a CI Job Token. This is something we do regularly in our CI/CD pipelines (e.g: upload/download something into the package registry). However, the HTTP header for passing the CI Job Token is not the same as a personal access token, which is why we're proposing the current change. I haven't done a thorough analysis of all the API calls. When the new Gitlab feature is added, I'll be able to test it to make sure it works end-to-end.

AlexandreEXFO:
@fgreinacher : I just tried 'curl --fail --header "JOB-TOKEN: $CI_JOB_TOKEN" "https://gitlab.com/api/v4/projects/<id>"' and it doesn't work from a pipeline. :(

fgreinacher:
Thanks for testing! We really need this to work with all the endpoints we're using to avoid frustration for users.

fgreinacher:
Let's go forward with #859 instead. Thanks for all the work @AlexandreEXFO!

Issue #822, opened 2025-02-14
Support releasing GitLab CI components


fgreinacher:
GitLab CI components need extra steps to be properly published to the catalog, see https://gitlab.com/gitlab-org/gitlab/-/issues/442066#note_1910121919 and more recently https://gitlab.com/gitlab-org/gitlab/-/issues/463253#note_2325647638.

xrow:
@fgreinacher This is true. /projects/:id/catalog/publish needs to be used if the project is a catalog project. I have the same issue.

PR #852, opened 2025-05-16
feat: raise error when GitLab CLI is required but not installed

fgreinacher:
Contributes to https://github.com/semantic-release/gitlab/issues/822

### DIFF
--- a/lib/glab.js
+++ b/lib/glab.js
@@ -1,0 +1,5 @@
+import { execa } from "execa";
+
+export default async (args, options) => {
+  return execa("glab", args, options);
+};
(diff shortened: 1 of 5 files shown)


lib/glab.js
fgreinacher, line 1: Moved this here to have a central place where we call the GitLab CI. In the future we could try downloading it here automatically.

lib/verify.js
```diff
@@ -89,6 +92,18 @@ export default async (pluginConfig, context) => {
     }
   }
 
+  if (publishToCatalog === true) {
```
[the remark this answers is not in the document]
fgreinacher: Let's keep it simple

lib/glab.js
JonasSchubert: **praise:** nice idea 👍

PR #860, opened 2025-06-06
feat: enable support for creating release tag via release api

arvest-bjoneson:
Enables creation of release tag via the `ref` field of the release api.

Implements a verification check to ensure all branches are set to use the default release channel (as notes cannot be pushed)

Depends on: https://github.com/semantic-release/semantic-release/pull/3760


arvest-bjoneson:
With [#3760 ](https://github.com/semantic-release/semantic-release/pull/3760) being rejected, this PR would be moot. Hoping maybe it will be reconsidered, but if not can close this PR as well. @travi

fgreinacher:
> With [#3760 ](https://github.com/semantic-release/semantic-release/pull/3760) being rejected, this PR would be moot. Hoping maybe it will be reconsidered, but if not can close this PR as well. @travi

I agree, let's close this one and focus on #859 instead.

PR #889, opened 2025-08-19
fix(generix-package): add missing fallback for label #823

### LINKED ISSUE #823: [Help] Gitlab release generic package is labeled undefined
I am trying to create a release, and upload an asset to the release that gets saved as a generic package on gitlab. I see in the readme that if the asset path is a glob that only finds one file it should use that file name as the generic package name. But this is not the case. This results in generic package that is my correct file, but it is named "undefined" with no file extension.

I have already verified with an exec plugin that there is in fact only a single .pdf file in my project, and it seems that the fact that the asset labeled undefined being the file contents I desired proves that case.

FYI - I am using `../*.pdf` for my path since my semrel config is stored within another directory.

Here is what my config looks like if you are curious:
```
plugins:
  # analyses the Git commits
  - - '@semantic-release/commit-analyzer'
    - preset: conventionalcommits
  # generates the release note from the Git commit messages
  - - '@semantic-release/release-notes-generator'
    - preset: conventionalcommits
  # generates the CHANGELOG.md file
  - - '@semantic-release/changelog'
    - changelogTitle: "# Changelog"
  # GitLab support
  - - '@semantic-release/gitlab'
    - assets: # save built pdf as generic package
      - path: "../*.pdf"
        target: "generic_package"

branches:
  - master
tagFormat: '${version}'

```

This seems to be a bug or incorrect functionality in documentation.

JonasSchubert:
This should fix #823 while adding the file basename fallback for a label as the description in the README highlights.

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -108,8 +108,9 @@ export default async (pluginConfig, context) => {
           let response;
 
           if (target === "generic_package") {
+            const finalLabel = label ?? pathlib.basename(file);
             // Upload generic packages
-            const encodedLabel = encodeURIComponent(label);
+            const encodedLabel = encodeURIComponent(finalLabel);
             // https://docs.gitlab.com/ee/user/packages/generic_packages/#publish-a-package-file
             uploadEndpoint = urlJoin(
               projectApiUrl,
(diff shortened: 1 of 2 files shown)


fgreinacher (approved):
Thanks, LGTM!

PR #901, opened 2025-09-10
Use CI JOB TOKEN to push code

tachyons:
<a href='https://hearts.dev/projects/3/users/tachyons'></a>

> This PR is built on top of https://github.com/semantic-release/gitlab/pull/859

GitLab now supports git push authenticted using CI JOB token. This PR update the authentication logic to also support CI JOB TOKEN when available.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+      - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           cache: npm
           node-version: lts/*
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       - name: "Use Node.js ${{ matrix.node-version }}"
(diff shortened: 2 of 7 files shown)


fgreinacher (commented):
<a href='https://hearts.dev/projects/3/users/fgreinacher'></a>

Sorry this took so long @tachyons.

I'm fine to go forward with the job token, but I'd like this to be a more conscious decision of the user so that they are not surprised by the limitations. What do you think about introducing a dedicated configuration option for "opting in" to the job token instead of just reusing the existing variable? We might also use the same option to disable unsupported features like issue/MR comments.

arvest-bjoneson:
<a href='https://hearts.dev/projects/3/users/arvest-bjoneson'></a>

@fgreinacher 

I might potentially argue that the user is already explicitly "opting in" by manually assigning either `GL_TOKEN` or `GITLAB_TOKEN` the value of the `CI_JOB_TOKEN`. The upstream semantic-release project already inherently supports this, as the git client does not require any special headers when using a job token.

IMHO, with the error messages and documentation changes introduced by the this PR, the user is made directly aware of the limitations (specifically around comments), and asked to explicitly disable them when using the job token.

I believe it may actually be more confusing to allow these other properties to be set, and disabled transitively by some other configuration flag like "useJobToken: true" or something.

At present, the code in the PR will automatically determine if a job token has been assigned (which again, a user must explicitly do), and invoke logic which provides clear feedback to the user on the limitations through error messages if they do not read the documentation.

It would be relatively trivial to add a boolean config property, like `useJobToken`, however the user would still be required to read error output and/or documentation to understand what the implications of doing so are. I believe such a property would be redundant to the signal the user is already providing by having to explicitly assign a job token to one of the existing token variables.

arvest-bjoneson:
<a href='https://hearts.dev/projects/3/users/arvest-bjoneson'></a>

@fgreinacher that sounds fair. If @tachyons is unable to respond, I'd be willing to take a crack at this over the next couple of days. Most of the changes are from a branch I had originally created.

PR #919, opened 2025-10-31
feat: enable use of job token for gitlab api authentication

arvest-bjoneson:
<a href='https://hearts.dev/projects/3/users/arvest-bjoneson'></a>

This is an implementation of recommendations made in https://github.com/semantic-release/gitlab/pull/901

PR #953, opened 2026-02-23
fix: header must be a string, not array

candrews:
<a href='https://hearts.dev/projects/3/users/candrews'></a>

The header must be a string, not an array. Per the got documentation, `headers` is of type `object<string, string>`. See https://github.com/sindresorhus/got/blob/v14.6.6/documentation/2-options.md#headers

This PR fixes the mis-typing problem introduced in https://github.com/semantic-release/gitlab/pull/919 which appears to cause GitLab releases to fail with incorrect `EINVALIDGLTOKEN Invalid GitLab token.` errors.

@arvest-bjoneson @fgreinacher


fgreinacher:
<a href='https://hearts.dev/projects/3/users/fgreinacher'></a>

@candrews The diff looks off. Please double check and ping me when ready for review.

